PR #4, opened 2017-01-23
Add basic  Wi-Fi API support.

### LINKED ISSUE #1: Add Wi-Fi support
Add support for basic Android Wi-Fi APIs.

xpconanfan:
* Create snippet class for WifiManager.
* Create JsonBuilder for serializing data types.
* Create a util for waiting on asynch operations.
* Enable Java8 for the project.
* Add formatter for the project.

Fixes #1


build.gradle
```diff
@@ -26,10 +30,22 @@ android {
...
 
 dependencies {
     compile 'com.google.android.mobly:mobly-snippet-lib:1.0.0'
     compile 'com.android.support.test:runner:0.5'
+    compile 'com.google.code.gson:gson:2.6.2'
 }
+
+googleJavaFormat {
+    options style: 'AOSP'
+}
```
adorokhine: github seems mad that there's no trailing newline here.

src/main/java/com/google/android/mobly/snippet/bundled/JsonBuilder.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    public JSONObject buildScanResult(ScanResult scanResult) throws JSONException {
+        return new JSONObject(mGson.toJson(scanResult));
+    }
+
+    public JSONObject buildWifiConfiguration(WifiConfiguration data) throws JSONException {
+        JSONObject result = new JSONObject(mGson.toJson(data));
+        String statueFieldName = "Status";
+        switch (data.status) {
```
adorokhine: Is this whole switch statement the same as result.put(statueFieldName, WifiConfiguration.Status.toString())?

src/main/java/com/google/android/mobly/snippet/bundled/JsonBuilder.java
```diff
@@ -0,0 +1,113 @@
...
+        return result;
+    }
+
+    public JSONObject buildScanResult(ScanResult scanResult) throws JSONException {
+        return new JSONObject(mGson.toJson(scanResult));
+    }
+
+    public JSONObject buildWifiConfiguration(WifiConfiguration data) throws JSONException {
+        JSONObject result = new JSONObject(mGson.toJson(data));
+        String statueFieldName = "Status";
```
adorokhine: 'statue'?

src/main/java/com/google/android/mobly/snippet/bundled/JsonBuilder.java
```diff
@@ -0,0 +1,113 @@
...
+                result.put(statueFieldName, "ENABLED");
+                break;
+            default:
+                result.put(statueFieldName, "UNKNOWN");
+        }
+        result.put("SSID", trimQuotationMarks(data.SSID));
+        return result;
+    }
+
+    public JSONObject buildWifiInfo(WifiInfo data) throws JSONException {
```
adorokhine: Mind adding javadoc to everything here that has none?

src/main/java/com/google/android/mobly/snippet/bundled/JsonBuilder.java
```diff
@@ -0,0 +1,113 @@
...
+import android.net.wifi.WifiInfo;
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
+import java.lang.reflect.Modifier;
+import java.net.InetAddress;
+import java.net.UnknownHostException;
+import org.json.JSONException;
+import org.json.JSONObject;
+
+public class JsonBuilder {
```
adorokhine: Can you add a javadoc to this class explaining what goes in here?
eg it looks like a dumping ground for any kind of generic json utils plus converters between android structs and json?
Should the latter go into some more specific class?

src/main/java/com/google/android/mobly/snippet/bundled/Utils.java
```diff
@@ -0,0 +1,48 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.android.mobly.snippet.bundled;
+
+public class Utils {
```
adorokhine: Should be final with private constructor.

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
```diff
@@ -0,0 +1,262 @@
...
+import java.util.concurrent.locks.ReentrantLock;
+import org.json.JSONArray;
+import org.json.JSONException;
+import org.json.JSONObject;
+
+/**
+ * Snippet class exposing Android APIs in WifiManager.
+ */
+public class WifiManagerSnippet implements Snippet {
+    private final Lock lock = new ReentrantLock();
```
adorokhine: No 'm' prefix on this one?

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
```diff
@@ -0,0 +1,262 @@
...
+    final Condition scanResultsAvailable = lock.newCondition();
```
adorokhine: Not private?

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
```diff
@@ -0,0 +1,262 @@
...
+    private WifiManager mWifiManager;
```
adorokhine: Everything that can be 'final' in the instance vars should be.

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
```diff
@@ -0,0 +1,262 @@
...
+    private Context mContext;
+    private WifiManager.WifiLock mWifiLock;
+    private String TAG = "WifiManagerSnippet";
+    private JsonBuilder mJsonBuilder = new JsonBuilder();
+    private boolean isScanning = false;
```
adorokhine: No m?

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
```diff
@@ -0,0 +1,262 @@
...
+
+    public WifiManagerSnippet() {
+        mContext = InstrumentationRegistry.getContext();
+        mWifiManager = (WifiManager) mContext.getSystemService(Context.WIFI_SERVICE);
+    }
+
+    public static WifiConfiguration jsonToWifiConfig(JSONObject j) throws JSONException {
```
adorokhine: Why is this public?

adorokhine: Can you add javadocs to the non-trivial methods?

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
```diff
@@ -0,0 +1,262 @@
...
+    public void shutdown() {
+        wifiLockRelease();
+    }
+
+    private static class WifiManagerSnippetException extends Exception {
+        public final String rpcMethod;
+
+        public WifiManagerSnippetException(String msg) {
+            super(msg);
+            this.rpcMethod = null;
```
adorokhine: The stack trace will contain this info (the rpc that was called) right? I'd rather not have to maintain this...

build.gradle
```diff
@@ -26,10 +30,22 @@ android {
...
 }
 
 dependencies {
     compile 'com.google.android.mobly:mobly-snippet-lib:1.0.0'
     compile 'com.android.support.test:runner:0.5'
+    compile 'com.google.code.gson:gson:2.6.2'
 }
+
+googleJavaFormat {
+    options style: 'AOSP'
```
adorokhine: Is there some way to configure github to show the right width corresponding to our chosen style? It's hard to review 100 char code at 80 chars.

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
```diff
@@ -0,0 +1,262 @@
...
+    }
+
+    @Rpc(description = "Turns on Wi-Fi with a 30s timeout.")
+    public void wifiEnable() throws InterruptedException, WifiManagerSnippetException {
+        if (!mWifiManager.setWifiEnabled(true)) {
+            throw new WifiManagerSnippetException(
+                    "wifiEnable", "Failed to initiate enabling Wi-Fi.");
+        }
+        Utils.Predicate waitCondition =
+                () -> mWifiManager.getWifiState() != WifiManager.WIFI_STATE_ENABLED;
```
adorokhine: Is this backwards?
You're waiting for the condition "wifiState != enabled" to become true?

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
```diff
@@ -0,0 +1,262 @@
...
+    public ArrayList<JSONObject> wifiGetConfiguredNetworks() throws JSONException {
+        ArrayList<JSONObject> networks = new ArrayList<>();
+        for (WifiConfiguration config : mWifiManager.getConfiguredNetworks()) {
+            networks.add(mJsonBuilder.buildWifiConfiguration(config));
+        }
+        return networks;
+    }
+
+    @Rpc(description = "Acquires the Wi-Fi lock in full mode.")
+    public void wifiLockAcquireFull() {
```
adorokhine: Mind explaining the difference between full and scan-only mode?
When might a user want to choose one over the other?

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
```diff
@@ -0,0 +1,262 @@
...
+        }
+    }
+
+    @Rpc(description = "Get the information about the active Wi-Fi connection.")
+    public JSONObject wifiGetConnectionInfo() throws JSONException {
+        return mJsonBuilder.buildWifiInfo(mWifiManager.getConnectionInfo());
+    }
+
+    @Rpc(description = "Get the info from last successful DHCP request.")
+    public JSONObject wifiGetDhcpInfo() throws JSONException {
```
adorokhine: Are all these RPCs useful? I'd suggest it might be nice to add RPCs on a YAGNI basis so we have to maintain less code :)

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
```diff
@@ -0,0 +1,262 @@
...
+            results.put(mJsonBuilder.buildScanResult(result));
+        }
+        return results;
+    }
+
+    @Rpc(description = "Start scan, wait for scan to complete, and return results.")
+    public JSONArray wifiScanAndGetResults()
+            throws InterruptedException, JSONException, WifiManagerSnippetException {
+        mContext.registerReceiver(
+                new WifiScanReceiver(),
```
adorokhine: Do we ever need to unregister this?

build.gradle
xpconanfan: I don't know

src/main/java/com/google/android/mobly/snippet/bundled/JsonBuilder.java
xpconanfan: status is an int

src/main/java/com/google/android/mobly/snippet/bundled/JsonBuilder.java
xpconanfan: what does 'has none' mean?

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
xpconanfan: what's the convention for using 'm'?

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
xpconanfan: In case other classes need to use it in the future.

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
xpconanfan: the receiver unregisters itself.

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
xpconanfan: Don't intend to repeat documentation on `WifiManager.WIFI_MODE_*`.
I could add a link to the official Android documentation, but even that seem redundant since anybody who cares enough to use this call should already know what it means or should be capable of looking up Android doc

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
xpconanfan: getDhcpInfo is pretty useful if you need to figure out connection info, since users don't have sys priv to access all the goodies in connectivity manager here.

Maybe the lock ones are not as useful

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
xpconanfan: Done

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
xpconanfan: No, waitCondition defines the condition for keep waiting.
So here it means when the state is not 'Enabled', we want to keep waiting.

src/main/java/com/google/android/mobly/snippet/bundled/JsonBuilder.java
adorokhine: Okay... then is this the same as

    result.put(statueFieldName, WifiConfiguration.Status.strings[data.status])?

src/main/java/com/google/android/mobly/snippet/bundled/JsonBuilder.java
adorokhine: Mind adding javadoc here to all public methods that currently have no javadoc?

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
adorokhine: I thought all member variables started with m per this style guide?

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
```diff
@@ -0,0 +1,262 @@
...
+/**
+ * Snippet class exposing Android APIs in WifiManager.
+ */
+public class WifiManagerSnippet implements Snippet {
+    private final Lock lock = new ReentrantLock();
+    final Condition scanResultsAvailable = lock.newCondition();
+    private WifiManager mWifiManager;
+    private Context mContext;
+    private WifiManager.WifiLock mWifiLock;
+    private String TAG = "WifiManagerSnippet";
```
adorokhine: If this is a log tag, it should be a constant ("private static final String") instead of a member var.

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
adorokhine: YAGNI?

adorokhine: Agreed on the RPC methods. Let's do the longer utility methods.

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
adorokhine: Hmm, that kind of API seems dangerous.
For example I checked WifiManager and noticed there's a state called WIFI_STATE_DISABLING.
Per the logic of this function, it will return and think it's successful as soon as it leaves WIFI_STATE_ENABLED, so it'll return too soon and cause race conditions.
Could we make this a positive check (waitUntil) to avoid issues like this?

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
adorokhine: I'm not suggesting it's not useful, I'm only suggesting to add them when requested to try and keep the API a little smaller.

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
```diff
@@ -178,17 +169,17 @@ public void wifiRemoveNetwork(Integer networkId) throws WifiManagerSnippetExcept
         return networks;
     }
 
-    @Rpc(description = "Acquires the Wi-Fi lock in full mode.")
+    @Rpc(description = "Acquires the Wi-Fi mReentrantLock in full mode.")
```
adorokhine: Bad search and replace?
In eclipse and intellij ctrl+shift+r will do a context-aware rename so this doesn't happen...

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
```diff
@@ -0,0 +1,231 @@
...
+            mWifiLock.acquire();
+        }
+    }
+
+    @Override
+    public void shutdown() {
+        wifiLockRelease();
+    }
+
+    private static class WifiManagerSnippetException extends Exception {
```
adorokhine: Could we make the class names consistent between this and bluetooth?
(No preference for which convention we choose)

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
```diff
@@ -0,0 +1,231 @@
...
+    @Rpc(description = "Return a list of all the configured wifi networks.")
+    public ArrayList<JSONObject> wifiGetConfiguredNetworks() throws JSONException {
+        ArrayList<JSONObject> networks = new ArrayList<>();
+        for (WifiConfiguration config : mWifiManager.getConfiguredNetworks()) {
+            networks.add(mJsonBuilder.buildWifiConfiguration(config));
+        }
+        return networks;
+    }
+
+    @Rpc(description = "Acquires the Wi-Fi mReentrantLock in full mode.")
```
adorokhine: Bad search and replace?
In eclipse and intellij alt+shift+r does a context-aware variable rename to avoid things like this...
(or right click -> refactor -> rename)

src/main/java/com/google/android/mobly/snippet/bundled/JsonBuilder.java
xpconanfan: Seems redundant, they are just all gonna say `serializing Xxx type to json`, adding no value but more strings to maintain...

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
xpconanfan: Well, similar methods are actually used by other classes in sl4a implementations.

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
xpconanfan: This was actually done by Android Studio's refactor, so, :/
Done

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
xpconanfan: I'm actually using this myself to check connection status as I develop.
So I think keeping it is fine.
We might be ok removing the wifi lock stuff

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
xpconanfan: Sure, what does the style guide say about the name though?
In Python, it was sufficient to just call it 'Error', can we do something similar here to reduce name length or is it different in Java.

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
xpconanfan: But, there is no util methods in this class other than the short `acquireWifiLock`...

build.gradle
xpconanfan: Actually searching on google suggests that it depends on your browser...

src/main/java/com/google/android/mobly/snippet/bundled/JsonBuilder.java
adorokhine: Bump

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
adorokhine: Bump

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
adorokhine: When those classes appear here we can flip the method from private to public. Should be pretty easy.
Also I thought utils dealing with Json were in JsonFormatter?

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
adorokhine: Bump

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
adorokhine: Sure, let's remove what you think we don't need.

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
adorokhine: I'm not sure, I haven't read the aosp style guide. Mind taking a look?

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
```diff
@@ -45,7 +45,7 @@
     private WifiManager.WifiLock mWifiLock;
     private final String TAG = "WifiManagerSnippet";
     private final JsonBuilder mJsonBuilder = new JsonBuilder();
-    private boolean isScanning = false;
+    private boolean mIsScanning = false;
```
adorokhine: If this could be accessed from multiple threads, needs to be marked 'volatile'.

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
xpconanfan: What is `JsonFormatter`

adorokhine: Sorry, I meant JsonBuilder.

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
xpconanfan: I was thinking the exact same case yet somehow I put `!` at the wrong place...
Done

src/main/java/com/google/android/mobly/snippet/bundled/JsonBuilder.java
xpconanfan: Done

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
xpconanfan: which style guide are we following... [this](https://source.android.com/source/code-style.html)?

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
xpconanfan: JsonBuilder is meant to serialize a Java data obj, this is the opposite of that.

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
xpconanfan: Don't see anything related to this
https://source.android.com/source/code-style.html

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
xpconanfan: Done

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
adorokhine: I don't know, I thought you were picking?
What does "options style: 'AOSP'" in our build file correspond to?

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
adorokhine: Ah I see. So methods to go from struct to json to into JsonBuilder, and methods to go from json to struct live in the snippet most relevant to that struct? Is that the convention we are following? Seems a little bit inconsistent...

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
adorokhine: There's no difference between a != b and !(a == b). The problem still exists.

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
adorokhine: Bump

src/main/java/com/google/android/mobly/snippet/bundled/JsonBuilder.java
xpconanfan: Done
Not sure why this is still showing old code...

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
xpconanfan: Ah, that's for the formatter plugin.
It can do google style as well. We just need to pick one

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
xpconanfan: It's the 'convention' we followed internally.
Maybe they should be named `XxxSerializer` and `XxxDeserializer`.
What prefix to use though, 'Rpc'?

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
xpconanfan: ughhhh, what was i thinking...

xpconanfan: Wait, I don't get it, why isn't the original logic sufficient?
It will wait as long as `state is not Enabled`, so it won't return when state is `Enabling`.

adorokhine: Yep, you're totally right, not enough sleep I guess...
Don't suppose we could flip the logic to make it a condition to wait for, not a condition to wait while? And rename the method waitFor or waitUntil?
I think it might be a little easier to brain.
waitFor(wifiState == ENABLED)
vs
waitAndCheck(wifiState != ENABLED)

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
adorokhine: JsonSerializer and JsonDeserializer maybe?

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
xpconanfan: I think that's what I meant when I said I put the '!' in the wrong place...
Didn't get much sleep on the ski trip... ╚(ಠ_ಠ)=┐

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
adorokhine: Agreed, there's nothing about naming exceptions. So let's make a convention and stick to it.
One of the problems with Java is that usually only the simple class name is printed, not the package or path, so both wifi_snippet.Error and bluetooth_snippet.Error would just say "Error:" iirc. I guess you'd need to test though.
If that's true then WifiSnippetException and BluetoothSnippetException sgtm?

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
adorokhine: Bump

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
```diff
@@ -71,7 +71,7 @@ public void wifiEnable() throws InterruptedException, WifiManagerSnippetExceptio
             throw new WifiManagerSnippetException("Failed to initiate enabling Wi-Fi.");
         }
         Utils.Predicate waitCondition =
-                () -> mWifiManager.getWifiState() != WifiManager.WIFI_STATE_ENABLED;
+                () -> !(mWifiManager.getWifiState() == WifiManager.WIFI_STATE_ENABLED);
```
adorokhine: Could you put these back then?

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
```diff
@@ -224,7 +224,7 @@ public void onReceive(Context c, Intent intent) {
             String action = intent.getAction();
             if (action.equals(WifiManager.SCAN_RESULTS_AVAILABLE_ACTION)) {
                 mIsScanning = false;
-                scanResultsAvailable.signal();
+                mScanResultsAvailable.signal();
             }
         }
     }
```
adorokhine: Were you going to delete those lock methods?

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
xpconanfan: Done

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
xpconanfan: Done

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
xpconanfan: Done

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
xpconanfan: One exception class per snippet class
Done

xpconanfan: Done

src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java
```diff
@@ -0,0 +1,116 @@
...
+    private String trimQuotationMarks(String originalString) {
+        String result = originalString;
+        if (originalString.charAt(0) == '"'
+                && originalString.charAt(originalString.length() - 1) == '"') {
+            result = originalString.substring(1, originalString.length() - 1);
+        }
+        return result;
+    }
+
+    public JSONObject toJson(Object object) throws JSONException {
```
xpconanfan: should we make everything else in this class private?

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
```diff
@@ -0,0 +1,196 @@
...
+        Utils.Predicate expectedState = () -> mIsScanResultAvailable;
+        if (!Utils.waitUntil(expectedState, 2 * 60)) {
+            throw new WifiManagerSnippetException(
+                    "Failed to get scan results after 2min, timeout!");
+        }
+        return wifiGetScanResults();
+    }
+
+    @Rpc(description = "Connects to a Wi-Fi network.")
+    public void wifiConnect(JSONObject wifiNetworkConfig)
```
adorokhine: Any reason this has to be a jsonobject?
eg why not just make it an SSID and @Nullable password?
If it has to be a jsonobject for some reason, the contents of the object should definitely be described to the user.

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
```diff
@@ -0,0 +1,196 @@
...
+        }
+        Utils.Predicate expectedState =
+                () -> mWifiManager.getConnectionInfo().getSSID().isEmpty();
+        if (!Utils.waitUntil(expectedState, 30)) {
+            throw new WifiManagerSnippetException("Failed to disconnect Wi-Fi after 30s, timeout!");
+        }
+    }
+
+    @Rpc(description = "Forget a configured Wi-Fi network by its network ID.")
+    public void wifiRemoveNetwork(Integer networkId) throws WifiManagerSnippetException {
```
adorokhine: 1) Why is networkId Integer instead of int?
2) Is it useful to a user to be able to remove wifi networks by network id? Wouldn't SSID be more user-friendly?
3) Do we need this call? I don't think it has any users...

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
```diff
@@ -0,0 +1,196 @@
...
+        if (!mWifiManager.removeNetwork(networkId)) {
+            throw new WifiManagerSnippetException("Failed to remove network of ID: " + networkId);
+        }
+    }
+
+    @Rpc(description = "Return a list of all the configured wifi networks.")
+    public ArrayList<JSONObject> wifiGetConfiguredNetworks() throws JSONException {
```
adorokhine: Should describe what JSONObject contains, ie that it will contain serialized WifiConfiguration objects

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
```diff
@@ -0,0 +1,196 @@
...
+        ArrayList<JSONObject> networks = new ArrayList<>();
+        for (WifiConfiguration config : mWifiManager.getConfiguredNetworks()) {
+            networks.add(mJsonSerializer.toJson(config));
+        }
+        return networks;
+    }
+
+    @Rpc(description = "Get the information about the active Wi-Fi connection.")
+    public JSONObject wifiGetConnectionInfo() throws JSONException {
```
adorokhine: Should describe what the json object contains

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
```diff
@@ -0,0 +1,196 @@
...
+        return mJsonSerializer.toJson(mWifiManager.getConnectionInfo());
+    }
+
+    @Rpc(description = "Get the info from last successful DHCP request.")
+    public JSONObject wifiGetDhcpInfo() throws JSONException {
```
adorokhine: Should describe what the json object contains

src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java
adorokhine: sgtm

src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java
```diff
@@ -0,0 +1,116 @@
...
+        if (object instanceof DhcpInfo) {
+            return serializeDhcpInfo((DhcpInfo) object);
+        } else if (object instanceof ScanResult) {
+            return serializeScanResult((ScanResult) object);
```
adorokhine: Seems to do the same thing as the default case?

src/main/java/com/google/android/mobly/snippet/bundled/utils/Utils.java
```diff
@@ -0,0 +1,53 @@
...
+     * @param predicate A lambda function that specifies the condition to wait for. This function
+     *     should return true when the desired state has been reached.
+     * @param timeout The number of seconds to wait for before giving up.
+     * @return true if the operation finished before timeout, false otherwise.
+     * @throws InterruptedException
+     */
+    public static boolean waitUntil(Utils.Predicate predicate, int timeout)
+            throws InterruptedException {
+        while (!predicate.waitCondition() && timeout >= 0) {
+            Thread.sleep(1000);
```
adorokhine: How about we make this a little faster? Say 100ms?

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
xpconanfan: Because network config can have many different things and you can connect to wifi using many different ways. A fixed signature won't work here.
Will add a javadoc

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
xpconanfan: 1. i thought params have to be packed type to work for Rpc.
2. no, some wifi networks dont' have a fixed SSID
3. if there's add, there should be remove. these are basics. and tests will need to remove networks they added to properly clean up

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
xpconanfan: Can we bring back the returns field in @Rpc for this?
I'd rather document there than the string which is less likely to be looked at.

src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java
xpconanfan: Done

src/main/java/com/google/android/mobly/snippet/bundled/utils/Utils.java
xpconanfan: Done

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
xpconanfan: Even though the deserializer doesn't do the other fields today, it would be good to keep the signature open for extension than having to change it later.

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
xpconanfan: Even though the deserializer doesn't do the other fields today, it would be good to keep the signature open for extension than having to change it later.

adorokhine: Can you give me an example of what other fields might come up in the future that would matter for this call?

I thought ssid and password would be pretty universal for connecting to wifi...

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
adorokhine: How would wifi networks without an ssid be added in the first place? The add call takes an ssid... any reason they shouldn't be consistent?

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
adorokhine: Maybe... file an FR?
Any reason you don't want to put the contents of the returned struct into the rpc description? That's already what it contains, just without the description of the contents.

xpconanfan: Because then there's no way to enforce a consistent formatting of the 'return' and 'params' part of description string.

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
xpconanfan: It's called a passpoint network and it is becoming a requirement for more and more parties.
I'll add it when it's required, but I'd want to leave the signature open so we don't have to change it later.
http://www.wi-fi.org/discover-wi-fi/wi-fi-certified-passpoint

adorokhine: Hm, interesting. However this seems like a less common case. Would it perhaps make sense to have wifiConnect() be the simple call, and if users exceed its capabilities, they can call something more specific like wifiConnectPasspoint() or maybe wifiConnectByInfo() or something?

xpconanfan: I don't think it is too much to ask people to write calls like this:
`ad.api.wifiConnect({'SSID': 'GoogleGuest'})`

And it saves the trouble of having multiple connection APIs for now.

adorokhine: We don't need to have multiple connection APIs now. We only need to have it when we start getting people who need this functionality, which might be never.

It makes a very simple and common API confusing to use without examples/looking at the code, for a niche and not currently supported feature...

It makes help() useless too (unless you start documenting the params in the RPC description).

Also by the time we support these network types default args will be working again so we can have the same call support both without a dict, but just with a native python call.

I really don't like the weird dict api. :/

xpconanfan: Unfortunately it is demanded by the reality of Android APIs.

Passpoint is just an example, it is growing fast and we need to consider potential users like carriers and OEMs, who will need such features. And there are other popular Wi-Fi connection types like WPS etc. Just because we don't use it in our labs doesn't mean they are 'niche'. Having worked on Wi-Fi for some time gives me some perspective on this.

We can't duplicate all the fields of WifiConfiguration into the signature so the Json input is the cleanest way for now, unless we want to start stubbing data classes on the Python side, which Autotest did.

adorokhine: Are we targeting carriers and OEMs primarily? I thought the fact that we're not system signed pretty much precludes most of that usage. I thought our target audience was users who just need to be able to do simple things with their system to get their mobly tests to work, like wifi connect/disconnect, bt connect/disconnect, airplane mode on/off, etc...

And for more advanced users we can definitely have a separate advanced call that takes the full json. Best of both worlds!

Agreed that it would be painful to duplicate everything in WifiConfiguration in the method args if there's a lot.

xpconanfan: `ad.api.wifiConnect({'SSID': 'GoogleGuest'})` doesn't seem too much for regular user to type and it is kinda nice since it mimics kwargs. So why not just have one API instead of many.

If we split, it'll be confusing to understand exactly what wifiConnect does - it connects to some networks but not others, by an arbitrary standard.

Or we'd have to be really explicit and have `wifiConnectWpa`, `wifiConnectWpa2`, `wifiConnectWps` etc.

All considered, having one wifiConnect seems the better way.

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
dthkao: I think the open signature is useful but maybe as a separate RPC (wifiConnectByConfig). For most people unfamiliar with the Android wifi the point is to make sure their test device is connected to their test network.

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
adorokhine: Are wpa, wpa2, and wps not all covered by username and password?
I don't think it's confusing: it connects to any network that accepts a username and optional password. Which is 99.9% of them.

xpconanfan: It is definitely NOT 99.9% of them.
WPS have three different connection modes itself, and they are not 'password' based.
Then there are the RSA based ones which are quite common.

adorokhine: I've never in my life seen an AP that wasn't open, WEP, WPA or WPA2 (except complex enterprise ones requiring certificates to be installed on the device and other steps). I suspect this is the case for most user environments as well.

We should definitely support it as an option, but I definitely don't think requiring users to go through the most complex, most generic flow in all cases is user-friendly or valuable. It'll just make it a hard-to-use API for functionality most people don't need.

Honestly I don't understand the problem with having a simple and advanced call. Seems pretty straightfoward to me.

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
```diff
@@ -0,0 +1,216 @@
...
+
+    @Rpc(description = "Trigger Wi-Fi scan.")
+    public void wifiStartScan() throws WifiManagerSnippetException {
+        if (!mWifiManager.startScan()) {
+            throw new WifiManagerSnippetException("Failed to initiate Wi-Fi scan.");
+        }
+    }
+
+    @Rpc(description = "Get Wi-Fi scan results.")
+    public JSONArray wifiGetScanResults() throws JSONException {
```
adorokhine: Will this RPC work without registering a receiver?
Seems like the only way to use this properly is through wifiScanAndGetResults() ?

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+    @Rpc(description = "Disconnect current Wi-Fi connection.")
+    public void wifiDisconnect() throws InterruptedException, WifiManagerSnippetException {
+        if (!mWifiManager.disconnect()) {
+            throw new WifiManagerSnippetException("Failed to initiate disconnecting Wi-Fi.");
+        }
+        // Disconnecting from current network doesn't necessarily mean the end state is no network
+        // connected to, because the device will try to roam to one of the saved networks, possibly
+        // the same one we just disconnected from.
```
adorokhine: Not sure how to interpret this comment... does this mean even after disconnection the wifi might reconnect to the same AP anyway? Under what circumstances would disconnect be used then?

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+    @Rpc(description = "Forget a configured Wi-Fi network by its network ID.")
+    public void wifiRemoveNetwork(Integer networkId) throws WifiManagerSnippetException {
```
adorokhine: How do you get a network ID? Does it come from wifiGetConfiguredNetworks?

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
adorokhine: Could you still stick it in the javadoc in the meantime?

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
```diff
@@ -0,0 +1,216 @@
...
+     *
+     * @param wifiNetworkConfig A JSON object that contains the info required to connect to a Wi-Fi
+     *     network. It follows the fields of WifiConfiguration type, e.g. {"SSID": "myWifi",
+     *     "password": "12345678"}.
+     * @throws InterruptedException
+     * @throws JSONException
+     * @throws WifiManagerSnippetException
+     */
+    @Rpc(description = "Connects to a Wi-Fi network.")
+    private void wifiConnect(JSONObject wifiNetworkConfig)
```
adorokhine: Can you call a private @Rpc method? I haven't tested but reflection might not be able to find it...

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
xpconanfan: Yes, the phone will always try to reconnect to any saved network.
Disconnect can be used to interrupt the connection. Or when a network has too low an RSSI, actively disconnecting Wi-Fi forces the device to cellular data.

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
xpconanfan: Yes.

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
xpconanfan: I actually didn't mean to make this private.
Fixed.

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
xpconanfan: Yes it works.
You can start scan, sleep for a while, and call this.

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
xpconanfan: I just realized that there are way too many fields to document since it's auto serialized by Gson.
Also the exact fields depend on the Android version has the data type changed over time.

src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java
xpconanfan: Done

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
adorokhine: We shouldn't be encouraging people to sleep in their tests so I don't think that use case is a good idea...
Retrieving cached results sgtm if you think it's valuable, but can we add a comment saying this RPC is to retrieve cached results and scanning should be done through wifiScanAndGetResults()?

Alternatively we can make this a private method until someone approaches us with the use case of re-retrieving cached scan results :) seems like if they want that they should just cache them on the python side...

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
adorokhine: How do you tell when disconnect is 'safe' (ie when the rssi is low enough that it won't try to reconnect)?
Is the intended use case of this API that someone does a scan, determines if RSSI is below some threshold, and if so calls disconnect?
Or is the intended use case that someone removes a saved network, and THEN calls disconnect, knowing that it's safe to do so since it won't try to reconnect to it anymore?
I think we need better documentation for how our API is supposed to be used...

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
adorokhine: sg, can you mention that it returns a serialized wifiConfiguration object and let people figure it out from there?
Should we at least document the wifi 'id' parameter that's used for our saved network removal API? So that people know how to call it? (can be either here or in the remove call docs)

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
xpconanfan: I was just giving some examples on how this API could be used, and the first example is still valid.
The intended use case is whenever the user needs to call WifiManager.disconnect.
We could speculate when users would do that, but we wouldn't be very successful. We can't think of all the use cases of users, but that doesn't mean this is not useful?

This is almost a oneliner wrap around the API from WifiManager, and I don't think copying paste docs from actual Android classes is a good idea.

adorokhine: We need to decide if we're exposing the raw android API or not. The reason that wifi_test_utils, tel_test_utils, etc sprouted up is because the raw android API is not user-friendly or usable for testing purposes. But the existence of such utils tie the user to a particular language and framework and is hard to keep backwards-compatible.

Basically I want to take what used to be in tel_test_utils and wifi_test_utils, and instead make a snippet out of them. So in this case yes, we should be carefully considering what API we want to expose and how to best wrap it into convenient, logical chunks for users.

Otherwise we need to write this snippet differently because right now we have a lot of convenience logic and wrappers. We can't straddle the fence.

xpconanfan: Ok, so we should take out the disconnect?
Since there really is no disconnect in android any more...
The real disconnect these days is 'disable'...

adorokhine: Taking it out and letting users use 'disable' sgtm, unless/until we have a reasonable use case and a way to call it reliably and deterministically.

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
xpconanfan: Sometimes I just want to get scan results without doing a whole new scan.
Also in manual debugging, I might trigger a scan via Settings, then get the results from snippet cli client.
This one is actually useful.

adorokhine: sgtm, should we throw 'cached' into the name to make it clearer that calling this method won't do a scan? getCachedScanResults?

xpconanfan: Oh myy, do we get right all the mistakes in Android API design here?! :P

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
xpconanfan: Done

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
xpconanfan: Done

xpconanfan:
Formatter did weird things with the annotation string...

README.md
```diff
@@ -11,7 +11,7 @@ Note: this is not an official Google product.
 1.  Compile and install the bundled snippets
 
         ./gradlew assembleDebug
-        adb install -d -r ./build/outputs/apk/mobly-bundled-snippets-debug.apk
+        adb install -d -r -r ./build/outputs/apk/mobly-bundled-snippets-debug.apk
```
adorokhine: Two -r flags?

xpconanfan: Done

README.md
```diff
@@ -21,15 +21,16 @@ Note: this is not an official Google product.
...
 
     ```python
     def setup_class(self):
       self.ad = self.register_controllers(android_device, min_number=1)[0]
       self.ad.load_snippet('api', 'com.google.android.mobly.snippet.bundled')
 
-    def test_enable_bluetooth(self):
-      self.ad.api.bluetoothEnable()
+    def test_enable_wifi(self):
+      self.ad.api.wifiEnable()
```
adorokhine: Why did we change this example from bluetooth to wifi?
They should both be equally well supported by us...
also now the method call is not consistent with the help() fragment in step 2.

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
```diff
@@ -0,0 +1,226 @@
...
+        }
+        return wifiGetCachedScanResults();
+    }
+
+    @Rpc(
+        description =
+                "Connects to a Wi-Fi network. This covers the common network types like "
+                        + "open and WPA2."
+    )
+    public void wifiConnectSimple(String SSID, @Nullable String password)
```
adorokhine: Per the style guide, the name of this variable should be ssid, not SSID:
https://source.android.com/source/code-style.html#treat-acronyms-as-words

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
```diff
@@ -0,0 +1,226 @@
...
+        }
+        Log.d(
+                "Connected to network '"
+                        + wifiConfig.SSID
+                        + "' with ID "
+                        + mWifiManager.getConnectionInfo().getNetworkId());
+    }
+
+    @Rpc(description = "Forget a configured Wi-Fi network by its network ID.")
+    public void wifiRemoveNetwork(Integer networkId) throws WifiManagerSnippetException {
```
adorokhine: Can you mention in the comment that this networkId is available in wifiConfiguration.networkId?

README.md
xpconanfan: Because we are going to have a more functional wifi snippet than bt.
will fix the help

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
xpconanfan: Done

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
xpconanfan: Done

README.md
```diff
@@ -18,7 +20,10 @@ Note: this is not an official Google product.
         snippet_shell.py com.google.android.mobly.snippet.bundled
         >>> print(s.help())
         Known methods:
-          bluetoothDisable() returns void  // Enable bluetooth
+          bluetoothDisable() returns void  // Disable bluetooth with a 30s timeout.
+        ...
+          wifiDisable() returns void  // Turns on Wi-Fi with a 30s timeout.
```
adorokhine: Both wifiDisable() and wifiEnable() say "turns *on* wi-fi"

xpconanfan: Done

adorokhine: This doc still says "turns on" for both so README.md still needs an update.

xpconanfan: Sorry ;(

PR #7, opened 2017-01-28
Add target for source jar.

adorokhine:


adorokhine:
Importing this artefact into third-party requires checking in both the classes jar and the sources jar. This is much more difficult to do if the build system has no way to produce one.

adorokhine:
Done.

PR #11, opened 2017-02-03
Stop building with Jack.

adorokhine:
jack doesn't produce a classes.jar and is slower than the old toolchain.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -4,6 +4,7 @@ buildscript {
     }
     dependencies {
         classpath 'com.android.tools.build:gradle:2.2.3'
+        classpath 'me.tatarka:gradle-retrolambda:3.5.0'
 
         // NOTE: Do not place your application dependencies here.
     }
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -18,6 +19,7 @@ repositories {
 }
 
 apply plugin: 'com.android.application'
+apply plugin: 'me.tatarka.retrolambda'
```
xpconanfan: hmm, does this mean we don't get Java8 features other than lambda?

adorokhine:
Which Java 8 features are you interested in?

xpconanfan:
We use "Type Annotations" as well right? The `@Nullable` for example.
For other features, it would be nice if we could keep the options open as well.

adorokhine:
Type annotations means that annotations can now be used in 'instanceof', 'new', and casts. Annotations could always be used at the point of declaration, which is what @Nullable is. So @Nullable will continue to work.

There aren't any other features jack gives us that retrolambda doesn't.

PR #13, opened 2017-02-09
Add makeToast.

### LINKED ISSUE #9: Add makeToast
Add an rpc that makes a toast on the screen. This serves as an easy way to make sure the bundled snippet functions.

xpconanfan:
fixes #9

### DIFF
--- a/src/main/java/com/google/android/mobly/snippet/bundled/NotificationSnippet.java
+++ b/src/main/java/com/google/android/mobly/snippet/bundled/NotificationSnippet.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright (C) 2017 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/android/mobly/snippet/bundled/NotificationSnippet.java
```diff
@@ -0,0 +1,48 @@
...
+     */
+    private final Handler mHandler;
+
+    public NotificationSnippet() {
+        mContext = InstrumentationRegistry.getContext();
+        mHandler = new Handler(mContext.getMainLooper());
+    }
+
+    @Rpc(description = "Make a toast on screen.")
+    public void makeToast(String message) {
```
adorokhine: In keeping with our convention, should we call this notificationMakeToast?
I'm ok either way.

xpconanfan: That seems very long...
I have mixed feelings about the prefix.
It was nice to dump things that we didn't want to categorize into `AndroidFacade` back then...
So shall we keep this one simple?...

PR #15, opened 2017-02-11
Target sdk 22 so we don't have to deal with runtime permissions.

adorokhine:
-g avoids them but users have to remember to specify that, plus Gradle
warns you that you are calling things in an invalid way unless you request
permissions through the runtime system.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -28,7 +28,8 @@ android {
     defaultConfig {
         applicationId "com.google.android.mobly.snippet.bundled"
         minSdkVersion 11
-        targetSdkVersion 24
+        // Set target to 22 to avoid having to deal with runtime permissions.
+        targetSdkVersion 22
         versionCode 1
         versionName "0.0.1"
         setProperty("archivesBaseName", "mobly-bundled-snippets")


xpconanfan:
Do we know how long will this trick last us?
I assume at one point they would force people to move up sdk level?
Also doesn't this stop us from using APIs added after 22?

adorokhine:
> Do we know how long will this trick last us? I assume at one point they would force people to move up sdk level?

No one forces you to move up sdk levels. You can still target sdk 1 if you want.

> Also doesn't this stop us from using APIs added after 22?

Nope, doesn't seem to.

xpconanfan:
Interesting, what's the down side then?
Why don't all apps just target API 22 and never move up?...

adorokhine:
Behaviour changes that come in through targetsdkversion are very visible to end-users so there is sales/marketing pressure to keep up. For example, apps targeting <22 installed from the Play store will throw up the big permissions acceptance dialog that the user needs to scroll through and accept. Apps targeting 23+ will be immediately installed without having to accept permissions.

Since our app isn't in the play store, this isn't an issue for us.

More here:
https://medium.com/google-developers/picking-your-compilesdkversion-minsdkversion-targetsdkversion-a098a0341ebd#.grdvv26fc

PR #16, opened 2017-02-11
Support adding and listing Google accounts.

adorokhine:
This change is [Reviewable](https://reviewable.io/reviews/google/mobly-bundled-snippets/16)

### DIFF
--- a/src/main/java/com/google/android/mobly/snippet/bundled/AccountSnippet.java
+++ b/src/main/java/com/google/android/mobly/snippet/bundled/AccountSnippet.java
@@ -67,7 +67,7 @@ public class AccountSnippet implements Snippet {
         mSyncStatusObserverHandles = new LinkedList<>();
     }
 
-    @Rpc(description = "Add a Google account to the device.")
+    @Rpc(description = "Add a Google account to the device, with account data sync disabled.")
     public void addGoogleAccount(String username, String password)
         throws AccountSnippetException, AccountsException, IOException {
         Bundle addAccountOptions = new Bundle();
--- a/src/main/java/com/google/android/mobly/snippet/bundled/AccountSnippet.java
+++ b/src/main/java/com/google/android/mobly/snippet/bundled/AccountSnippet.java
@@ -88,7 +88,7 @@ public class AccountSnippet implements Snippet {
         // a little hard to understand.
         if (listAccounts().contains(username)) {
             throw new AccountSnippetException(
-                "Account " + username + " alreaady exists on the device");
+                "Account " + username + " already exists on the device");
         }
         Bundle addAccountOptions = new Bundle();
         addAccountOptions.putString("username", username);
--- a/src/main/java/com/google/android/mobly/snippet/bundled/AccountSnippet.java
+++ b/src/main/java/com/google/android/mobly/snippet/bundled/AccountSnippet.java
@@ -1,0 +1,115 @@
+/*
+ * Copyright (C) 2017 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.android.mobly.snippet.bundled;
+
+import android.accounts.Account;
+import android.accounts.AccountManager;
+import android.accounts.AccountManagerCallback;
+import android.accounts.AccountManagerFuture;
+import android.accounts.AccountsException;
+import android.accounts.AuthenticatorException;
+import android.accounts.OperationCanceledException;
+import android.content.ContentResolver;
+import android.content.Context;
+import android.content.SyncAdapterType;
+import android.content.SyncStatusObserver;
+import android.os.Bundle;
+import android.os.Handler;
+import android.support.test.InstrumentationRegistry;
+import android.widget.Toast;
+import com.google.android.mobly.snippet.Snippet;
+import com.google.android.mobly.snippet.rpc.Rpc;
+import com.google.android.mobly.snippet.util.Log;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+/** Snippet class exposing Android APIs related to creating notification on screen. */
+public class AccountSnippet implements Snippet {
+    private static final String GOOGLE_ACCOUNT_TYPE = "com.google";
+    private static final String AUTH_TOKEN_TYPE = "mail";
+
+    private static class AccountSnippetException extends Exception {
+        public AccountSnippetException(String msg) {
+            super(msg);
+        }
+    }
+
+    private final AccountManager mAccountManager;
+
+    public AccountSnippet() {
+        Context context = InstrumentationRegistry.getContext();
+        mAccountManager = AccountManager.get(context);
+    }
+
+    @Rpc(description = "Add a Google account to the device.")
+    public void addGoogleAccount(String username, String password)
+        throws AccountSnippetException, AccountsException, IOException {
+        Bundle addAccountOptions = new Bundle();
+        addAccountOptions.putString("username", username);
+        addAccountOptions.putString("password", password);
+        AccountManagerFuture<Bundle> future =
+            mAccountManager.addAccount(
+                GOOGLE_ACCOUNT_TYPE,
+                AUTH_TOKEN_TYPE,
+                null /* requiredFeatures */,
+                addAccountOptions,
+                null /* activity */,
+                null /* authCallback */,
+                null /* handler */);
+        Bundle result = future.getResult();
+        if (result.containsKey(AccountManager.KEY_ERROR_CODE)) {
+            throw new AccountSnippetException(
+                String.format("Failed to add account due to code %d: %s",
+                    result.getInt(AccountManager.KEY_ERROR_CODE),
+                    result.getString(AccountManager.KEY_ERROR_MESSAGE)));
+        }
+
+        // Disable sync to avoid test flakiness as accounts fetch additional data.
+        // It takes a while for all sync adapters to be populated, so register for broadcasts when
+        // sync is starting and disable them there.
+        // NOTE: this listener is NOT unregistered because several sync requests for the new account
+        // will come in over time.
+        Account account = new Account(username, GOOGLE_ACCOUNT_TYPE);
+        ContentResolver.addStatusChangeListener(
+            ContentResolver.SYNC_OBSERVER_TYPE_ACTIVE | ContentResolver.SYNC_OBSERVER_TYPE_PENDING,
+            which -> {
+                Log.i("Attempt to sync account " + username + " detected! Disabling.");
+                for (SyncAdapterType adapter : ContentResolver.getSyncAdapterTypes()) {
+                    if (!adapter.accountType.equals(GOOGLE_ACCOUNT_TYPE)) {
+                        continue;
+                    }
+                    ContentResolver
+                        .setSyncAutomatically(account, adapter.authority, false /* sync */);
+                    ContentResolver.cancelSync(account, adapter.authority);
+                }
+            });
+    }
+
+    @Rpc(description = "List all Google accounts on the device.")
+    public List<String> listGoogleAccounts() throws SecurityException {
+        Account[] accounts = mAccountManager.getAccountsByType(GOOGLE_ACCOUNT_TYPE);
+        List<String> usernames = new ArrayList<>(accounts.length);
+        for (Account account : accounts) {
+            usernames.add(account.name);
+        }
+        return usernames;
+    }
+
+    @Override
+    public void shutdown() {}
+}
--- a/src/main/java/com/google/android/mobly/snippet/bundled/AccountSnippet.java
+++ b/src/main/java/com/google/android/mobly/snippet/bundled/AccountSnippet.java
@@ -36,8 +36,11 @@ import com.google.android.mobly.snippet.rpc.Rpc;
 import com.google.android.mobly.snippet.util.Log;
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Set;
+import java.util.TreeSet;
 
 /**
  * Snippet class exposing Android APIs related to management of device accounts.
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/android/mobly/snippet/bundled/AccountSnippet.java
```diff
@@ -0,0 +1,115 @@
...
+        mAccountManager = AccountManager.get(context);
+    }
+
+    @Rpc(description = "Add a Google account to the device.")
+    public void addGoogleAccount(String username, String password)
+        throws AccountSnippetException, AccountsException, IOException {
+        Bundle addAccountOptions = new Bundle();
+        addAccountOptions.putString("username", username);
+        addAccountOptions.putString("password", password);
+        AccountManagerFuture<Bundle> future =
```
xpconanfan: could/should the `.getResult()` just be called directly instead of creating the `future` var.

src/main/java/com/google/android/mobly/snippet/bundled/AccountSnippet.java
```diff
@@ -0,0 +1,115 @@
...
+                AUTH_TOKEN_TYPE,
+                null /* requiredFeatures */,
+                addAccountOptions,
+                null /* activity */,
+                null /* authCallback */,
+                null /* handler */);
+        Bundle result = future.getResult();
+        if (result.containsKey(AccountManager.KEY_ERROR_CODE)) {
+            throw new AccountSnippetException(
+                String.format("Failed to add account due to code %d: %s",
```
xpconanfan: What's our convention on string formatting?
It seems that most Android code simply use `+`.
I'm open to anything as long as it's consistent :)

src/main/java/com/google/android/mobly/snippet/bundled/AccountSnippet.java
```diff
@@ -0,0 +1,115 @@
...
+import android.widget.Toast;
+import com.google.android.mobly.snippet.Snippet;
+import com.google.android.mobly.snippet.rpc.Rpc;
+import com.google.android.mobly.snippet.util.Log;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+/** Snippet class exposing Android APIs related to creating notification on screen. */
+public class AccountSnippet implements Snippet {
```
xpconanfan: Can we have a javadoc for this class explaining what this does?
e.g. What is a google account, what other types of accounts there are.

src/main/java/com/google/android/mobly/snippet/bundled/AccountSnippet.java
```diff
@@ -0,0 +1,115 @@
...
+                null /* handler */);
+        Bundle result = future.getResult();
+        if (result.containsKey(AccountManager.KEY_ERROR_CODE)) {
+            throw new AccountSnippetException(
+                String.format("Failed to add account due to code %d: %s",
+                    result.getInt(AccountManager.KEY_ERROR_CODE),
+                    result.getString(AccountManager.KEY_ERROR_MESSAGE)));
+        }
+
+        // Disable sync to avoid test flakiness as accounts fetch additional data.
```
xpconanfan: Is this because if syncing is in progress, attempts to operate with the added account will fail?
Or if there's no connectivity, sync will fail anyway?

Can we have an Rpc that syncs the account and waits until the account is ready/sync finishes?

src/main/java/com/google/android/mobly/snippet/bundled/AccountSnippet.java
```diff
@@ -0,0 +1,115 @@
...
+        Account[] accounts = mAccountManager.getAccountsByType(GOOGLE_ACCOUNT_TYPE);
+        List<String> usernames = new ArrayList<>(accounts.length);
+        for (Account account : accounts) {
+            usernames.add(account.name);
+        }
+        return usernames;
+    }
+
+    @Override
+    public void shutdown() {}
```
xpconanfan: should the StatusChangeListener be unregistered here?

src/main/java/com/google/android/mobly/snippet/bundled/AccountSnippet.java
```diff
@@ -0,0 +1,115 @@
...
+            throw new AccountSnippetException(
+                String.format("Failed to add account due to code %d: %s",
+                    result.getInt(AccountManager.KEY_ERROR_CODE),
+                    result.getString(AccountManager.KEY_ERROR_MESSAGE)));
+        }
+
+        // Disable sync to avoid test flakiness as accounts fetch additional data.
+        // It takes a while for all sync adapters to be populated, so register for broadcasts when
+        // sync is starting and disable them there.
+        // NOTE: this listener is NOT unregistered because several sync requests for the new account
```
xpconanfan: Should we give user the option to not register this listener or an extra Rpc to unregister this?
If I want to simulate a real world use case, I'd probably want to sync the account?

src/main/java/com/google/android/mobly/snippet/bundled/AccountSnippet.java
```diff
@@ -0,0 +1,115 @@
...
+import android.support.test.InstrumentationRegistry;
+import android.widget.Toast;
+import com.google.android.mobly.snippet.Snippet;
+import com.google.android.mobly.snippet.rpc.Rpc;
+import com.google.android.mobly.snippet.util.Log;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+/** Snippet class exposing Android APIs related to creating notification on screen. */
```
adorokhine: Done

src/main/java/com/google/android/mobly/snippet/bundled/AccountSnippet.java
adorokhine: Done

src/main/java/com/google/android/mobly/snippet/bundled/AccountSnippet.java
```diff
@@ -0,0 +1,115 @@
...
+    private static class AccountSnippetException extends Exception {
+        public AccountSnippetException(String msg) {
+            super(msg);
+        }
+    }
+
+    private final AccountManager mAccountManager;
+
+    public AccountSnippet() {
+        Context context = InstrumentationRegistry.getContext();
```
adorokhine: It's not a question of variable naming. The 'm' prefix means member (instance variables). Local variables don't need a prefix.

If/when context is needed by multiple methods, it would become a member variable, and would be renamed.

src/main/java/com/google/android/mobly/snippet/bundled/AccountSnippet.java
adorokhine: It helps with step debugging to keep steps separate. eg stepping into addAccount and making sure the arguments are handled properly, then let it block on getResult() and switch to a different thread to make sure the async operation is handled ok. What's the advantage to doing things inline?

src/main/java/com/google/android/mobly/snippet/bundled/AccountSnippet.java
adorokhine: They're not equivalent. + is implemented as a StringBuilder so it's much faster but with awkward syntax when mixing template and vars, and no support for formatting like float rounding and zero padding. String.format() is slower but has a much easier syntax for mixing variables into the template and the ability to format the values.

In this case I didn't feel like doing `"due to code " + code + ": " + reason`; too many quotes and plusses, and it's not in a tight loop, so String.format() seemed like the better choice in this case.

src/main/java/com/google/android/mobly/snippet/bundled/AccountSnippet.java
adorokhine: > Is this because if syncing is in progress, attempts to operate with the added account will fail?
Or if there's no connectivity, sync will fail anyway?

No, it's to avoid flakiness:
- Apps owned by the account will start being fetched, potentially killing/replacing existing apps
- The phone will slow down with background sync operations, eg downloading gmails
- The account state will start changing in unpredictable ways at any time. eg backed up app data will start being downloaded, so app state for various apps will start changing.
- As saved wifi access points are downloaded, the phone might switch itself to a previously remembered wifi AP
- etc.

> Can we have an Rpc that syncs the account and waits until the account is ready/sync finishes?

Sadly there's no way to detect when sync is complete as far as I can see.

I'm happy to make an option to allow sync once kwargs support is in, but for now account sync should probably be disabled for test purposes. (and this option would be inherently racy for the reasons outlined above).

src/main/java/com/google/android/mobly/snippet/bundled/AccountSnippet.java
adorokhine: Mentioned above: I'm happy to make an option to allow sync once kwargs support is in, but for now account sync should probably be disabled for test purposes. (and this option would be inherently racy for the reasons outlined above).

> If I want to simulate a real world use case, I'd probably want to sync the account?

If there was a way to do this reliably, yes, but there's no way to detect that a sync is done so tests using such functionality would probably be flaky.

src/main/java/com/google/android/mobly/snippet/bundled/AccountSnippet.java
adorokhine: Done.

src/main/java/com/google/android/mobly/snippet/bundled/AccountSnippet.java
xpconanfan: Less code to read...
We don't lose any debug info if `.getResults` call is on it's own line, right?
Nbd though.

adorokhine: In my opinion it's the same amount of code to read, but harder to read it because each line does a lot of things. :)

Shorter, simpler lines are way easier for me to mentally process.
Plus it helps with debugging like I said.

src/main/java/com/google/android/mobly/snippet/bundled/AccountSnippet.java
xpconanfan: Could `ContentResolver.isSyncActive` and `ContentResolver.isSyncPending` be helpful here?
Also, let's mention the cancellation of sync for these accounts in the help string?

src/main/java/com/google/android/mobly/snippet/bundled/AccountSnippet.java
xpconanfan: So we'll allow both?

adorokhine: I would say yes, both are fine where each makes sense, taking into the account the performance versus convenience tradeoffs.

src/main/java/com/google/android/mobly/snippet/bundled/AccountSnippet.java
adorokhine: Sadly no. The listener I have actually registers for the same active and pending events that these functions return. Accounts will go into pending and active sync states many times, asynchronously, as new sync providers pick up the account and begin their activity. As far as I can see there's no way to know when it's permanently done.

src/main/java/com/google/android/mobly/snippet/bundled/AccountSnippet.java
```diff
@@ -67,8 +67,9 @@ public AccountSnippet() {
         mSyncStatusObserverHandles = new LinkedList<>();
     }
 
-    @Rpc(description = "Add a Google account to the device, with account data sync disabled.")
-    public void addGoogleAccount(String username, String password)
+    @Rpc(description =
+        "Add a Google (GMail) account to the device, with account data sync disabled.")
+    public void addAccount(String username, String password)
```
xpconanfan: why the rename?
The old names seem better since we are specifically using `GOOGLE_ACCOUNT_TYPE` in the method.

adorokhine: Per the commit message:

>  This will let us add a 'type' kwarg for other account types once kwargs are supported.

xpconanfan: Can you add a TODO in the code, so we don't lose track of this reasoning.

src/main/java/com/google/android/mobly/snippet/bundled/AccountSnippet.java
```diff
@@ -0,0 +1,146 @@
...
+     *
+     * <p>TODO(adorokhine): Support adding accounts of other types with an optional 'type' kwarg.
+     * <p>TODO(adorokhine): Allow users to choose whether to enable/disable sync with a kwarg.
+     *
+     * @param username Username of the account to add (including @gmail.com).
+     * @param password Password of the account to add.
+     */
+    @Rpc(description =
+        "Add a Google (GMail) account to the device, with account data sync disabled.")
+    public void addAccount(String username, String password)
```
xpconanfan: Since we can't remove account, what happens if `addAccount` code is called once, then snippet exits and restarts, then `addAccount` is called again with the same info? Would it just override?

adorokhine: Right now you get an exception from the Android API:

    android.accounts.AuthenticatorException: Account does not exist or not visible. Maybe change pwd?

So right now you'd have to list the accounts and avoid re-adding an account already on the phone.

I could work around this by doing an implicit lookup inside addAccount() if you wanted.

xpconanfan: It would be nice if the error message is accurate. :)

adorokhine: Agreed, but this is the error Android gives you. Our only option would be to catch it and wrap it with our own exception. Thoughts?

xpconanfan: Let's make it accurate with our own message then?

adorokhine: Done.

src/main/java/com/google/android/mobly/snippet/bundled/AccountSnippet.java
```diff
@@ -92,7 +96,17 @@ public void addAccount(String username, String password)
                 null /* activity */,
                 null /* authCallback */,
                 null /* handler */);
-        Bundle result = future.getResult();
+        Bundle result;
+        try {
+            result = future.getResult();
+        } catch (AuthenticatorException e) {
+            if (e.getMessage().equals("Account does not exist or not visible. Maybe change pwd?")) {
```
xpconanfan: Hmm, I thought the solution was to iterate through the list of existing accounts and throw this error.
Is that not feasible?

adorokhine: Done.

src/main/java/com/google/android/mobly/snippet/bundled/AccountSnippet.java
```diff
@@ -84,6 +83,13 @@ public AccountSnippet() {
         "Add a Google (GMail) account to the device, with account data sync disabled.")
     public void addAccount(String username, String password)
         throws AccountSnippetException, AccountsException, IOException {
+        // Check for existing account. If we try to re-add an existing account, Android throws an
+        // exception that says "Account does not exist or not visible. Maybe change pwd?" which is
+        // a little hard to understand.
+        if (listAccounts().contains(username)) {
+            throw new AccountSnippetException(
+                "Account " + username + " alreaady exists on the device");
```
xpconanfan: 'alreaady'

PR #19, opened 2017-02-14
Minor fixes to Mobly Bundled Snippets.

adorokhine:
* Remove unused imports
* Explicitly add the INTERNET permission, which is required for snippet
  to run (it came in implicitly through manfiest merge with
  mobly-snippet-lib but it's good to be explicit).

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly-bundled-snippets/19)


src/main/AndroidManifest.xml
```diff
@@ -10,6 +10,7 @@
     <uses-permission android:name="android.permission.BLUETOOTH_ADMIN" />
     <uses-permission android:name="android.permission.CHANGE_WIFI_STATE" />
     <uses-permission android:name="android.permission.GET_ACCOUNTS" />
+    <uses-permission android:name="android.permission.INTERNET" />
```
xpconanfan: I don't think this is needed?

PR #26, opened 2017-03-05
adding methords for BluetoothAdapterSnippet with serializer

ygw11223:
Four methods added:
public void bluetoothStartDiscovery()
public JSONArray bluetoothGetCachedScanResults()
public JSONArray bluetoothDiscoveryAndGetResults() // it calls the above two methods
public JSONArray bluetoothGetPairedDevices()
One class added:
private class BluetoothScanReceiver extends BroadcastReceiver{} // used for discovery
One serializer added:
private JSONObject serializeBtDevice(BluetoothDevice data)

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly-bundled-snippets/26)

### DIFF
--- a/src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java
+++ b/src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java
@@ -21,11 +21,14 @@ import android.net.DhcpInfo;
 import android.net.wifi.SupplicantState;
 import android.net.wifi.WifiConfiguration;
 import android.net.wifi.WifiInfo;
+import android.os.ParcelUuid;
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 import java.lang.reflect.Modifier;
 import java.net.InetAddress;
 import java.net.UnknownHostException;
+import java.util.ArrayList;
+import org.json.JSONArray;
 import org.json.JSONException;
 import org.json.JSONObject;
 
@@ -122,8 +125,45 @@ public class JsonSerializer {
     }
 
     private JSONObject serializeBluetoothDevice(BluetoothDevice data) throws JSONException {
-        JSONObject result = new JSONObject(mGson.toJson(data));
-        result.put("Name", data.getName());
+        JSONObject result = new JSONObject();
+        result.put("Address", data.getAddress());
+        switch (data.getBondState()) {
+            case 10:
+                result.put("BondState", "None");
+                break;
+            case 11:
+                result.put("BondState", "Bonding");
+                break;
+            case 12:
+                result.put("BondState", "Bonded");
+        }
+        if (data.getName() == null) {
+            result.put("Name", "Null");
+        } else {
+            result.put("Name", data.getName());
+        }
+        switch (data.getType()) {
+            case 0:
+                result.put("Type", "Unknown");
+                break;
+            case 1:
+                result.put("Type", "Classic");
+                break;
+            case 2:
+                result.put("Type", "Le");
+                break;
+            case 3:
+                result.put("Type", "Dual");
+        }
+        if (data.getUuids() == null) {
+            result.put("UUIDs", "Null");
+        } else {
+            ArrayList<String> uuids = new ArrayList<String>();
+            for (ParcelUuid uuid : data.getUuids()) {
+                uuids.add(uuid.toString());
+            }
+            result.put("UUIDs", new JSONArray(uuids));
+        }
         return result;
     }
 }
--- a/src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java
+++ b/src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java
@@ -69,7 +69,7 @@ public class JsonSerializer {
         } else if (object instanceof WifiInfo) {
             return serializeWifiInfo((WifiInfo) object);
         } else if (object instanceof BluetoothDevice) {
-            return serializeBtDevice((BluetoothDevice) object);
+            return serializeBluetoothDevice((BluetoothDevice) object);
         }
         return defaultSerialization(object);
     }
@@ -121,7 +121,7 @@ public class JsonSerializer {
         return result;
     }
 
-    private JSONObject serializeBtDevice(BluetoothDevice data) throws JSONException {
+    private JSONObject serializeBluetoothDevice(BluetoothDevice data) throws JSONException {
         JSONObject result = new JSONObject(mGson.toJson(data));
         result.put("NAME", data.getName());
         return result;
--- a/src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java
+++ b/src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java
@@ -20,6 +20,7 @@ import android.net.DhcpInfo;
 import android.net.wifi.SupplicantState;
 import android.net.wifi.WifiConfiguration;
 import android.net.wifi.WifiInfo;
+import android.bluetooth.BluetoothDevice;
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 import java.lang.reflect.Modifier;
@@ -67,6 +68,8 @@ public class JsonSerializer {
             return serializeWifiConfiguration((WifiConfiguration) object);
         } else if (object instanceof WifiInfo) {
             return serializeWifiInfo((WifiInfo) object);
+        } else if (object instanceof BluetoothDevice) {
+            return serializeBtDevice((BluetoothDevice) object);
         }
         return defaultSerialization(object);
     }
@@ -115,6 +118,12 @@ public class JsonSerializer {
                 result.put("SupplicantState", state.name());
             }
         }
+        return result;
+    }
+
+    private JSONObject serializeBtDevice(BluetoothDevice data) throws JSONException {
+        JSONObject result = new JSONObject(mGson.toJson(data));
+        result.put("NAME", data.getName());
         return result;
     }
 }
--- a/src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java
+++ b/src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java
@@ -62,14 +62,14 @@ public class JsonSerializer {
     }
 
     public JSONObject toJson(Object object) throws JSONException {
-        if (object instanceof DhcpInfo) {
+        if (object instanceof BluetoothDevice) {
+            return serializeBluetoothDevice((BluetoothDevice) object);
+        } else if (object instanceof DhcpInfo) {
             return serializeDhcpInfo((DhcpInfo) object);
         } else if (object instanceof WifiConfiguration) {
             return serializeWifiConfiguration((WifiConfiguration) object);
         } else if (object instanceof WifiInfo) {
             return serializeWifiInfo((WifiInfo) object);
-        } else if (object instanceof BluetoothDevice) {
-            return serializeBluetoothDevice((BluetoothDevice) object);
         }
         return defaultSerialization(object);
     }
@@ -123,7 +123,7 @@ public class JsonSerializer {
 
     private JSONObject serializeBluetoothDevice(BluetoothDevice data) throws JSONException {
         JSONObject result = new JSONObject(mGson.toJson(data));
-        result.put("NAME", data.getName());
+        result.put("Name", data.getName());
         return result;
     }
 }
--- a/src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java
+++ b/src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java
@@ -16,11 +16,11 @@
 
 package com.google.android.mobly.snippet.bundled.utils;
 
+import android.bluetooth.BluetoothDevice;
 import android.net.DhcpInfo;
 import android.net.wifi.SupplicantState;
 import android.net.wifi.WifiConfiguration;
 import android.net.wifi.WifiInfo;
-import android.bluetooth.BluetoothDevice;
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 import java.lang.reflect.Modifier;
--- a/src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java
+++ b/src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java
@@ -101,7 +101,7 @@ public class BluetoothAdapterSnippet implements Snippet {
         try {
             if (!mBluetoothAdapter.startDiscovery()) {
                 throw new BluetoothAdapterSnippetException(
-                        "Failed to initiate Bluetooth Discovery.");
+                    "Failed to initiate Bluetooth Discovery.");
             }
             if (!Utils.waitUntil(() -> mIsScanResultAvailable, 60)) {
                 throw new BluetoothAdapterSnippetException(
--- a/src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java
+++ b/src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java
@@ -42,7 +42,7 @@ public class BluetoothAdapterSnippet implements Snippet {
     private final Context mContext;
     private final BluetoothAdapter mBluetoothAdapter;
     private final JsonSerializer mJsonSerializer = new JsonSerializer();
-    private ArrayList<BluetoothDevice> mDiscoveryResults = new ArrayList<>();
+    private final ArrayList<BluetoothDevice> mDiscoveryResults = new ArrayList<>();
     private volatile boolean mIsScanResultAvailable = false;
 
     public BluetoothAdapterSnippet() {
@@ -70,8 +70,7 @@ public class BluetoothAdapterSnippet implements Snippet {
         }
     }
 
-    @Rpc(description = "Trigger bluetooth discovery.")
-    public void bluetoothStartDiscovery() throws BluetoothAdapterSnippetException {
+    private void bluetoothStartDiscovery() throws BluetoothAdapterSnippetException {
         if (!mBluetoothAdapter.isDiscovering()) {
             if (!mBluetoothAdapter.startDiscovery()) {
                 throw new BluetoothAdapterSnippetException(
@@ -108,13 +107,15 @@ public class BluetoothAdapterSnippet implements Snippet {
         mIsScanResultAvailable = false;
         BroadcastReceiver receiver = new BluetoothScanReceiver();
         mContext.registerReceiver(receiver, filter);
-        bluetoothStartDiscovery();
-        if (!Utils.waitUntil(() -> mIsScanResultAvailable, 60)) {
-            mContext.unregisterReceiver(receiver);
-            throw new BluetoothAdapterSnippetException(
+        try {
+            bluetoothStartDiscovery();
+            if (!Utils.waitUntil(() -> mIsScanResultAvailable, 60)) {
+                throw new BluetoothAdapterSnippetException(
                     "Failed to get discovery results after 1 min, timeout!");
+            }
+        } finally {
+            mContext.unregisterReceiver(receiver);
         }
-        mContext.unregisterReceiver(receiver);
         return bluetoothGetCachedScanResults();
     }
 
--- a/src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java
+++ b/src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java
@@ -43,7 +43,6 @@ public class BluetoothAdapterSnippet implements Snippet {
     private final BluetoothAdapter mBluetoothAdapter;
     private final JsonSerializer mJsonSerializer = new JsonSerializer();
     private ArrayList<BluetoothDevice> mDiscoveryResults = new ArrayList<>();
-    private volatile boolean mIsScanning = false;
     private volatile boolean mIsScanResultAvailable = false;
 
     public BluetoothAdapterSnippet() {
@@ -107,7 +106,6 @@ public class BluetoothAdapterSnippet implements Snippet {
         }
         mDiscoveryResults.clear();
         mIsScanResultAvailable = false;
-        mIsScanning = true;
         BroadcastReceiver receiver = new BluetoothScanReceiver();
         mContext.registerReceiver(receiver, filter);
         bluetoothStartDiscovery();
@@ -143,7 +141,6 @@ public class BluetoothAdapterSnippet implements Snippet {
         public void onReceive(Context context, Intent intent) {
             String action = intent.getAction();
             if (BluetoothAdapter.ACTION_DISCOVERY_FINISHED.equals(action)) {
-                mIsScanning = false;
                 mIsScanResultAvailable = true;
             } else if (BluetoothDevice.ACTION_FOUND.equals(action)) {
                 BluetoothDevice device =
--- a/src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java
+++ b/src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java
@@ -51,7 +51,7 @@ public class BluetoothAdapterSnippet implements Snippet {
     }
 
     @Rpc(description = "Enable bluetooth with a 30s timeout.")
-    public void bluetoothEnable() throws BluetoothAdapterSnippetException, InterruptedException {
+    public void btEnable() throws BluetoothAdapterSnippetException, InterruptedException {
         if (!mBluetoothAdapter.enable()) {
             throw new BluetoothAdapterSnippetException("Failed to start enabling bluetooth");
         }
@@ -61,7 +61,7 @@ public class BluetoothAdapterSnippet implements Snippet {
     }
 
     @Rpc(description = "Disable bluetooth with a 30s timeout.")
-    public void bluetoothDisable() throws BluetoothAdapterSnippetException, InterruptedException {
+    public void btDisable() throws BluetoothAdapterSnippetException, InterruptedException {
         if (!mBluetoothAdapter.disable()) {
             throw new BluetoothAdapterSnippetException("Failed to start disabling bluetooth");
         }
@@ -70,20 +70,11 @@ public class BluetoothAdapterSnippet implements Snippet {
         }
     }
 
-    private void bluetoothStartDiscovery() throws BluetoothAdapterSnippetException {
-        if (!mBluetoothAdapter.isDiscovering()) {
-            if (!mBluetoothAdapter.startDiscovery()) {
-                throw new BluetoothAdapterSnippetException(
-                        "Failed to initiate Bluetooth Discovery.");
-            }
-        }
-    }
-
     @Rpc(
         description =
                 "Get bluetooth discovery results, which is a list of serialized BluetoothDevice objects."
     )
-    public JSONArray bluetoothGetCachedScanResults() throws JSONException {
+    public JSONArray btGetCachedScanResults() throws JSONException {
         JSONArray results = new JSONArray();
         for (BluetoothDevice result : mDiscoveryResults) {
             results.put(mJsonSerializer.toJson(result));
@@ -96,7 +87,7 @@ public class BluetoothAdapterSnippet implements Snippet {
                 "Start discovery, wait for discovery to complete, and return results, which is a list of "
                         + "serialized BluetoothDevice objects."
     )
-    public JSONArray bluetoothDiscoveryAndGetResults()
+    public JSONArray btDiscoveryAndGetResults()
             throws InterruptedException, JSONException, BluetoothAdapterSnippetException {
         IntentFilter filter = new IntentFilter(BluetoothDevice.ACTION_FOUND);
         filter.addAction(BluetoothAdapter.ACTION_DISCOVERY_FINISHED);
@@ -108,7 +99,10 @@ public class BluetoothAdapterSnippet implements Snippet {
         BroadcastReceiver receiver = new BluetoothScanReceiver();
         mContext.registerReceiver(receiver, filter);
         try {
-            bluetoothStartDiscovery();
+            if (!mBluetoothAdapter.startDiscovery()) {
+                throw new BluetoothAdapterSnippetException(
+                        "Failed to initiate Bluetooth Discovery.");
+            }
             if (!Utils.waitUntil(() -> mIsScanResultAvailable, 60)) {
                 throw new BluetoothAdapterSnippetException(
                     "Failed to get discovery results after 1 min, timeout!");
@@ -116,11 +110,11 @@ public class BluetoothAdapterSnippet implements Snippet {
         } finally {
             mContext.unregisterReceiver(receiver);
         }
-        return bluetoothGetCachedScanResults();
+        return btGetCachedScanResults();
     }
 
     @Rpc(description = "Get the list of paired bluetooth devices")
-    public JSONArray bluetoothGetPairedDevices()
+    public JSONArray btGetPairedDevices()
             throws BluetoothAdapterSnippetException, InterruptedException, JSONException {
         JSONArray pairedDevices = new JSONArray();
         for (BluetoothDevice device : mBluetoothAdapter.getBondedDevices()) {
--- a/src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java
+++ b/src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java
@@ -17,11 +17,22 @@
 package com.google.android.mobly.snippet.bundled;
 
 import android.bluetooth.BluetoothAdapter;
+import android.bluetooth.BluetoothDevice;
 import android.content.Context;
+import android.content.Intent;
+import android.content.IntentFilter;
+import android.content.BroadcastReceiver;
 import android.support.test.InstrumentationRegistry;
 import com.google.android.mobly.snippet.Snippet;
+import com.google.android.mobly.snippet.bundled.utils.JsonDeserializer;
+import com.google.android.mobly.snippet.bundled.utils.JsonSerializer;
 import com.google.android.mobly.snippet.bundled.utils.Utils;
 import com.google.android.mobly.snippet.rpc.Rpc;
+import com.google.android.mobly.snippet.util.Log;
+import org.json.JSONArray;
+import org.json.JSONException;
+import org.json.JSONObject;
+import java.util.ArrayList;
 
 /** Snippet class exposing Android APIs in BluetoothAdapter. */
 public class BluetoothAdapterSnippet implements Snippet {
@@ -33,6 +44,10 @@ public class BluetoothAdapterSnippet implements Snippet {
 
     private final Context mContext;
     private final BluetoothAdapter mBluetoothAdapter;
+    private final JsonSerializer mJsonSerializer = new JsonSerializer();
+    private ArrayList<BluetoothDevice> mDiscoveryResults = new ArrayList<>();
+    private volatile boolean mIsScanning = false;
+    private volatile boolean mIsScanResultAvailable = false;
 
     public BluetoothAdapterSnippet() {
         mContext = InstrumentationRegistry.getContext();
@@ -59,6 +74,74 @@ public class BluetoothAdapterSnippet implements Snippet {
         }
     }
 
+    @Rpc(description = "Trigger bluetooth discovery.")
+    public void bluetoothStartDiscovery() throws BluetoothAdapterSnippetException {
+        if (mBluetoothAdapter.isDiscovering()) {
+            mBluetoothAdapter.cancelDiscovery();
+        }
+        if (!mBluetoothAdapter.startDiscovery()) {
+            throw new BluetoothAdapterSnippetException("Failed to initiate Bluetooth Discovery.");
+        }
+    }
+
+    @Rpc(
+        description =
+                "Get bluetooth discovery results, which is a list of serialized BluetoothDevice objects."
+    )
+    public JSONArray bluetoothGetCachedScanResults() throws JSONException {
+        JSONArray results = new JSONArray();
+        for (BluetoothDevice result : mDiscoveryResults) {
+            results.put(mJsonSerializer.toJson(result));
+        }
+        return results;
+    }
+
+    @Rpc(
+        description =
+                "Start discovery, wait for discovery to complete, and return results, which is a list of "
+                        + "serialized BluetoothDevice objects."
+    )
+    public JSONArray bluetoothDiscoveryAndGetResults()
+            throws InterruptedException, JSONException, BluetoothAdapterSnippetException {
+        IntentFilter filter = new IntentFilter(BluetoothDevice.ACTION_FOUND);
+        filter.addAction(BluetoothAdapter.ACTION_DISCOVERY_FINISHED);
+        BroadcastReceiver mReceiver = new BluetoothScanReceiver();
+        mContext.registerReceiver(mReceiver, filter);
+        mDiscoveryResults.clear();
+        bluetoothStartDiscovery();
+        mIsScanResultAvailable = false;
+        mIsScanning = true;
+        if (!Utils.waitUntil(() -> mIsScanResultAvailable, 60)) {
+            throw new BluetoothAdapterSnippetException(
+                    "Failed to get discovery results after 1 min, timeout!");
+        }
+        mContext.unregisterReceiver(mReceiver);
+        return bluetoothGetCachedScanResults();
+    }
+
+    @Rpc(description = "Get the list of paired bluetooth devices")
+    public JSONArray bluetoothGetPairedDevices() throws BluetoothAdapterSnippetException, InterruptedException, JSONException {
+        JSONArray pairedDevices = new JSONArray();
+        for (BluetoothDevice device : mBluetoothAdapter.getBondedDevices())
+            pairedDevices.put(mJsonSerializer.toJson(device));
+        return pairedDevices;
+    }
+
     @Override
     public void shutdown() {}
+
+    private class BluetoothScanReceiver extends BroadcastReceiver {
+
+        @Override
+        public void onReceive(Context c, Intent intent) {
+            String action = intent.getAction();
+            if (BluetoothAdapter.ACTION_DISCOVERY_FINISHED.equals(action)) {
+                mIsScanning = false;
+                mIsScanResultAvailable = true;
+            } else if (BluetoothDevice.ACTION_FOUND.equals(action)) {
+                BluetoothDevice device = (BluetoothDevice) intent.getParcelableExtra(BluetoothDevice.EXTRA_DEVICE);
+                mDiscoveryResults.add(device);
+            }
+        }
+    }
 }
--- a/src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java
+++ b/src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java
@@ -18,21 +18,18 @@ package com.google.android.mobly.snippet.bundled;
 
 import android.bluetooth.BluetoothAdapter;
 import android.bluetooth.BluetoothDevice;
+import android.content.BroadcastReceiver;
 import android.content.Context;
 import android.content.Intent;
 import android.content.IntentFilter;
-import android.content.BroadcastReceiver;
 import android.support.test.InstrumentationRegistry;
 import com.google.android.mobly.snippet.Snippet;
-import com.google.android.mobly.snippet.bundled.utils.JsonDeserializer;
 import com.google.android.mobly.snippet.bundled.utils.JsonSerializer;
 import com.google.android.mobly.snippet.bundled.utils.Utils;
 import com.google.android.mobly.snippet.rpc.Rpc;
-import com.google.android.mobly.snippet.util.Log;
+import java.util.ArrayList;
 import org.json.JSONArray;
 import org.json.JSONException;
-import org.json.JSONObject;
-import java.util.ArrayList;
 
 /** Snippet class exposing Android APIs in BluetoothAdapter. */
 public class BluetoothAdapterSnippet implements Snippet {
@@ -76,11 +73,11 @@ public class BluetoothAdapterSnippet implements Snippet {
 
     @Rpc(description = "Trigger bluetooth discovery.")
     public void bluetoothStartDiscovery() throws BluetoothAdapterSnippetException {
-        if (mBluetoothAdapter.isDiscovering()) {
-            mBluetoothAdapter.cancelDiscovery();
-        }
-        if (!mBluetoothAdapter.startDiscovery()) {
-            throw new BluetoothAdapterSnippetException("Failed to initiate Bluetooth Discovery.");
+        if (!mBluetoothAdapter.isDiscovering()) {
+            if (!mBluetoothAdapter.startDiscovery()) {
+                throw new BluetoothAdapterSnippetException(
+                        "Failed to initiate Bluetooth Discovery.");
+            }
         }
     }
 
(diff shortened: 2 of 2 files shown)


adorokhine:
Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, 13 unresolved discussions.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java, line 78 at r1](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-KeWH3ZflrStro20C4Gi:-KeWH3ZflrStro20C4Gj:bmyqpbn) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/1bfc5dcf49070354e0fa6e6aaa9d241a985c40f7/src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java#L78)):*
> ```Java
> 
>     @Rpc(description = "Trigger bluetooth discovery.")
>     public void bluetoothStartDiscovery() throws BluetoothAdapterSnippetException {
> ```

@Ang Li I thought we weren't doing the namespace prefixes anymore? I'm confused w.r.t our convention.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java, line 79 at r1](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-KeWHCckKvhxipcq2Tmq:-KeWHCckKvhxipcq2Tmr:b-9d1a7k) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/1bfc5dcf49070354e0fa6e6aaa9d241a985c40f7/src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java#L79)):*
> ```Java
>     @Rpc(description = "Trigger bluetooth discovery.")
>     public void bluetoothStartDiscovery() throws BluetoothAdapterSnippetException {
>         if (mBluetoothAdapter.isDiscovering()) {
> ```

Is there a reason we want to kill an existing discovery and re-start it, versus assuming it's already in the right state and doing nothing? Does this clear some state/scan with new settings/reset the timeout or something?

---

*[src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java, line 82 at r1](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-KeWHS6mUdRqNPhOMJSn:-KeWHS6mUdRqNPhOMJSo:by0b031) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/1bfc5dcf49070354e0fa6e6aaa9d241a985c40f7/src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java#L82)):*
> ```Java
>             mBluetoothAdapter.cancelDiscovery();
>         }
>         if (!mBluetoothAdapter.startDiscovery()) {
> ```

Does this persist forever or does it stop discovering eventually?

---

*[src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java, line 108 at r1](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-KeWIw0-n48o-u9vJW23:-KeWIw0-n48o-u9vJW24:byahy7l) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/1bfc5dcf49070354e0fa6e6aaa9d241a985c40f7/src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java#L108)):*
> ```Java
>         IntentFilter filter = new IntentFilter(BluetoothDevice.ACTION_FOUND);
>         filter.addAction(BluetoothAdapter.ACTION_DISCOVERY_FINISHED);
>         BroadcastReceiver mReceiver = new BluetoothScanReceiver();
> ```

Local variables shouldn't have the m prefix (m means member)

---

*[src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java, line 109 at r1](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-KeWIR6eWiCcpq_RVvos:-KeWIR6eWiCcpq_RVvot:b-fuh9rn) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/1bfc5dcf49070354e0fa6e6aaa9d241a985c40f7/src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java#L109)):*
> ```Java
>         filter.addAction(BluetoothAdapter.ACTION_DISCOVERY_FINISHED);
>         BroadcastReceiver mReceiver = new BluetoothScanReceiver();
>         mContext.registerReceiver(mReceiver, filter);
> ```

Do you need to make sure discovery is off already when you do this? Otherwise as soon as you register the receiver it could start getting results, right? Then mDiscoveryResults.clear() would delete those results?

---

*[src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java, line 112 at r1](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-KeWIdbuGMUoVAo1Cvnr:-KeWIdbvWqGJP-6uXnbm:b-j4pubk) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/1bfc5dcf49070354e0fa6e6aaa9d241a985c40f7/src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java#L112)):*
> ```Java
>         mDiscoveryResults.clear();
>         bluetoothStartDiscovery();
>         mIsScanResultAvailable = false;
> ```

Race condition here; after you start discovery it might finish at any time including before this line is executed. In that case you will end up waiting forever.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java, line 118 at r1](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-KeWIoSTrpqtMzytE4xJ:-KeWIoSTrpqtMzytE4xK:b-ks0kpb) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/1bfc5dcf49070354e0fa6e6aaa9d241a985c40f7/src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java#L118)):*
> ```Java
>                     "Failed to get discovery results after 1 min, timeout!");
>         }
>         mContext.unregisterReceiver(mReceiver);
> ```

Does this need to be in a finally block? eg it won't happen if you timeout?

---

*[src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java, line 125 at r1](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-KeWJ07AvVeJi7K01jaE:-KeWJ07EFlVWA-R3lGjU:b1dgvkc) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/1bfc5dcf49070354e0fa6e6aaa9d241a985c40f7/src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java#L125)):*
> ```Java
>     public JSONArray bluetoothGetPairedDevices() throws BluetoothAdapterSnippetException, InterruptedException, JSONException {
>         JSONArray pairedDevices = new JSONArray();
>         for (BluetoothDevice device : mBluetoothAdapter.getBondedDevices())
> ```

Curly braces even for single line if statements please

---

*[src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java, line 136 at r1](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-KeWHuc9BQGm9tOzTjYu:-KeWHucAoWAGTAxtDntb:brgr5d2) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/1bfc5dcf49070354e0fa6e6aaa9d241a985c40f7/src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java#L136)):*
> ```Java
> 
>         @Override
>         public void onReceive(Context c, Intent intent) {
> ```

No abbreviations except for loop variables please

---

*[src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java, line 138 at r1](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-KeWI159i-7L_2Iv6HPz:-KeWI159i-7L_2Iv6HQ-:b-n1u3na) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/1bfc5dcf49070354e0fa6e6aaa9d241a985c40f7/src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java#L138)):*
> ```Java
>         public void onReceive(Context c, Intent intent) {
>             String action = intent.getAction();
>             if (BluetoothAdapter.ACTION_DISCOVERY_FINISHED.equals(action)) {
> ```

Can you add a comment mentioning how this state machine works? eg does it give you a bunch of ACTION_FOUND events and then one final DISCOVERY_FINISHED?

---

*[src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java, line 23 at r1](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-KeWJB0SvVOdzxt2vWZu:-KeWJB0T-_RYcrp7BAnT:b-7euqh4) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/1bfc5dcf49070354e0fa6e6aaa9d241a985c40f7/src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java#L23)):*
> ```Java
> import android.net.wifi.WifiConfiguration;
> import android.net.wifi.WifiInfo;
> import android.bluetooth.BluetoothDevice;
> ```

Did you add this import manually? Not sorted in alphabetical order versus the other imports, but I always thought the IDE took care of that...

---

*[src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java, line 72 at r1](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-KeWJHRyok1S_rrPCyGI:-KeWJHRzJ9VMv-x6KI_e:b-ld5b3q) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/1bfc5dcf49070354e0fa6e6aaa9d241a985c40f7/src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java#L72)):*
> ```Java
>             return serializeWifiInfo((WifiInfo) object);
>         } else if (object instanceof BluetoothDevice) {
>             return serializeBtDevice((BluetoothDevice) object);
> ```

Mind avoiding abbreviations here?

---

*[src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java, line 126 at r1](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-KeWJLegiFmNAHv01p1I:-KeWJLegiFmNAHv01p1J:b-b6jdyy) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/1bfc5dcf49070354e0fa6e6aaa9d241a985c40f7/src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java#L126)):*
> ```Java
>     private JSONObject serializeBtDevice(BluetoothDevice data) throws JSONException {
>         JSONObject result = new JSONObject(mGson.toJson(data));
>         result.put("NAME", data.getName());
> ```

Does the name not end up in the returned object otherwise? I would have thought that normally gson would automatically take care of it unless name was a computed variable?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26)*

ygw11223:
@adorokhine Thank you for your valuable review! I've made some adjustments regarding your comments.

"Does the name not end up in the returned object otherwise? I would have thought that normally gson would automatically take care of it unless name was a computed variable?"
It turns out that gson builder only automatically serializes the mac address of the bluetooth device. I add the "NAME" field because it helps me to distinguish different devices on the shell side... Do you want me to remove it?

" if (!mBluetoothAdapter.startDiscovery()) { "
"Does this persist forever or does it stop discovering eventually?"
According to the API document, "If Bluetooth state is not STATE_ON, this API will return false." So I assume that this if condition is true only if the user initializes a discovery when bluetooth is off.

adorokhine:
Would you mind opening up this PR in [Reviewable](https://reviewable.io/reviews/google/mobly-bundled-snippets/26) and hitting 'done'/responding to the comments inline? It's a little hard to see what's been addressed so far/continue the discussions. :)

ygw11223:
Review status: 0 of 2 files reviewed at latest revision, 13 unresolved discussions.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java, line 79 at r1](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-KeWHCckKvhxipcq2Tmq:-Kg7BShCCm5_Bo_4fuUc:bybwgj2) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/1bfc5dcf49070354e0fa6e6aaa9d241a985c40f7/src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java#L79)):*
<details><summary><i>Previously, adorokhine (Alexander Dorokhine) wrote…</i></summary><blockquote>

Is there a reason we want to kill an existing discovery and re-start it, versus assuming it's already in the right state and doing nothing? Does this clear some state/scan with new settings/reset the timeout or something?
</blockquote></details>

Done. If there's a existing discovery, do nothing.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java, line 82 at r1](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-KeWHS6mUdRqNPhOMJSn:-Kg7CFmM6nejxSLZAxzp:b-giditm) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/1bfc5dcf49070354e0fa6e6aaa9d241a985c40f7/src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java#L82)):*
<details><summary><i>Previously, adorokhine (Alexander Dorokhine) wrote…</i></summary><blockquote>

Does this persist forever or does it stop discovering eventually?
</blockquote></details>

According to the API document, "If Bluetooth state is not STATE_ON, this [startDiscovery()] API will return false." So I assume that this if condition is true only if the user initializes a discovery when bluetooth is off.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java, line 108 at r1](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-KeWIw0-n48o-u9vJW23:-Kg7C_Q89uD-Q-DYHpDY:b-896fix) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/1bfc5dcf49070354e0fa6e6aaa9d241a985c40f7/src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java#L108)):*
<details><summary><i>Previously, adorokhine (Alexander Dorokhine) wrote…</i></summary><blockquote>

Local variables shouldn't have the m prefix (m means member)
</blockquote></details>

Done.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java, line 109 at r1](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-KeWIR6eWiCcpq_RVvos:-Kg7ClBDANc5XZXA_-Pz:bn1h5gr) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/1bfc5dcf49070354e0fa6e6aaa9d241a985c40f7/src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java#L109)):*
<details><summary><i>Previously, adorokhine (Alexander Dorokhine) wrote…</i></summary><blockquote>

Do you need to make sure discovery is off already when you do this? Otherwise as soon as you register the receiver it could start getting results, right? Then mDiscoveryResults.clear() would delete those results?
</blockquote></details>

Check added.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java, line 112 at r1](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-KeWIdbuGMUoVAo1Cvnr:-Kg7Cxyg-a45ls4SidCl:b87bv7h) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/1bfc5dcf49070354e0fa6e6aaa9d241a985c40f7/src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java#L112)):*
<details><summary><i>Previously, adorokhine (Alexander Dorokhine) wrote…</i></summary><blockquote>

Race condition here; after you start discovery it might finish at any time including before this line is executed. In that case you will end up waiting forever.
</blockquote></details>

Done by switching order of codes.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java, line 118 at r1](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-KeWIoSTrpqtMzytE4xJ:-Kg7DIHu4e2HcmUMDTbM:b7lzghc) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/1bfc5dcf49070354e0fa6e6aaa9d241a985c40f7/src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java#L118)):*
<details><summary><i>Previously, adorokhine (Alexander Dorokhine) wrote…</i></summary><blockquote>

Does this need to be in a finally block? eg it won't happen if you timeout?
</blockquote></details>

I added the unregister to the timeout block. Is it necessary to make a finally block?

---

*[src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java, line 125 at r1](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-KeWJ07AvVeJi7K01jaE:-Kg7DXQeOx4Xv7PjGFs3:b-896fix) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/1bfc5dcf49070354e0fa6e6aaa9d241a985c40f7/src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java#L125)):*
<details><summary><i>Previously, adorokhine (Alexander Dorokhine) wrote…</i></summary><blockquote>

Curly braces even for single line if statements please
</blockquote></details>

Done.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java, line 136 at r1](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-KeWHuc9BQGm9tOzTjYu:-Kg7D_AavIVBPrMq9nJP:b-896fix) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/1bfc5dcf49070354e0fa6e6aaa9d241a985c40f7/src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java#L136)):*
<details><summary><i>Previously, adorokhine (Alexander Dorokhine) wrote…</i></summary><blockquote>

No abbreviations except for loop variables please
</blockquote></details>

Done.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java, line 138 at r1](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-KeWI159i-7L_2Iv6HPz:-Kg7DdzJL22N1wYnwfK6:b-896fix) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/1bfc5dcf49070354e0fa6e6aaa9d241a985c40f7/src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java#L138)):*
<details><summary><i>Previously, adorokhine (Alexander Dorokhine) wrote…</i></summary><blockquote>

Can you add a comment mentioning how this state machine works? eg does it give you a bunch of ACTION_FOUND events and then one final DISCOVERY_FINISHED?
</blockquote></details>

Done.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java, line 23 at r1](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-KeWJB0SvVOdzxt2vWZu:-Kg7Dm0wyT16_ixWNXj2:b-896fix) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/1bfc5dcf49070354e0fa6e6aaa9d241a985c40f7/src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java#L23)):*
<details><summary><i>Previously, adorokhine (Alexander Dorokhine) wrote…</i></summary><blockquote>

Did you add this import manually? Not sorted in alphabetical order versus the other imports, but I always thought the IDE took care of that...
</blockquote></details>

Done.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java, line 72 at r1](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-KeWJHRyok1S_rrPCyGI:-Kg7Dyr9ywLVFTAa0Lm0:b-896fix) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/1bfc5dcf49070354e0fa6e6aaa9d241a985c40f7/src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java#L72)):*
<details><summary><i>Previously, adorokhine (Alexander Dorokhine) wrote…</i></summary><blockquote>

Mind avoiding abbreviations here?
</blockquote></details>

Done.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java, line 126 at r1](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-KeWJLegiFmNAHv01p1I:-Kg7E2Xa0xKbuxq9HeHe:bg8nym4) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/1bfc5dcf49070354e0fa6e6aaa9d241a985c40f7/src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java#L126)):*
<details><summary><i>Previously, adorokhine (Alexander Dorokhine) wrote…</i></summary><blockquote>

Does the name not end up in the returned object otherwise? I would have thought that normally gson would automatically take care of it unless name was a computed variable?
</blockquote></details>

It turns out that gson builder only automatically serializes the mac address of the bluetooth device. I add the "NAME" field because it helps me to distinguish different devices on the shell side... Do you want me to remove it?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26)*

ygw11223:
Done. Sorry about that...I am new to Reviewable.

---

Review status: 0 of 2 files reviewed at latest revision, 13 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-:-Kg7EZkKuMwZackZ6-C9:bs67hwc)*

adorokhine:
Reviewed 1 of 2 files at r2, 1 of 1 files at r3.
Review status: all files reviewed at latest revision, 6 unresolved discussions.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java, line 82 at r1](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-KeWHS6mUdRqNPhOMJSn:-KgMQ2FE8WCjd-U7ufOK:b-f4qa9b) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/1bfc5dcf49070354e0fa6e6aaa9d241a985c40f7/src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java#L82)):*
<details><summary><i>Previously, ygw11223 wrote…</i></summary><blockquote>

According to the API document, "If Bluetooth state is not STATE_ON, this [startDiscovery()] API will return false." So I assume that this if condition is true only if the user initializes a discovery when bluetooth is off.
</blockquote></details>

From the same API document, it looks like it persists for about 12 seconds. Given that, do we need to expose it as its own separate RPC? Can we make it a private method and let users use the bluetoothDiscoveryAndGetResults() RPC instead, unless you can think of a use case?

---

*[src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java, line 118 at r1](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-KeWIoSTrpqtMzytE4xJ:-KgMR-CiY44ve6uBZDFC:b-ruf9fj) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/1bfc5dcf49070354e0fa6e6aaa9d241a985c40f7/src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java#L118)):*
<details><summary><i>Previously, ygw11223 wrote…</i></summary><blockquote>

I added the unregister to the timeout block. Is it necessary to make a finally block?
</blockquote></details>

Yeah let's make a finally block. bluetoothStartDiscovery() can throw exceptions as well, and if that happens we'll leave the receive registered.

```
mContext.registerReceiver(receiver, filter);
try {
  bluetoothStartDiscovery();
  if (!Utils.waitUntil(() -> mIsScanResultAvailable, 60)) {
    throw new BluetoothAdapterSnippetException(
        "Failed to get discovery results after 1 min, timeout!");
  }
} finally {
  mContext.unregisterReceiver();
}
```


---

*[src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java, line 45 at r3](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-KgMRnWD0iiAG8PYLggg:-KgMRnWEKHQTcKlLIyqP:b-102yq6) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/ef831968684f98b748627cd0b4707eb0f868efe8/src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java#L45)):*
> mDiscoveryResults
Should be `final`.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java, line 126 at r1](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-KeWJLegiFmNAHv01p1I:-KgMQZljMoV1Gj0b7Hoq:b-csuh2y) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/1bfc5dcf49070354e0fa6e6aaa9d241a985c40f7/src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java#L126)):*
<details><summary><i>Previously, ygw11223 wrote…</i></summary><blockquote>

It turns out that gson builder only automatically serializes the mac address of the bluetooth device. I add the "NAME" field because it helps me to distinguish different devices on the shell side... Do you want me to remove it?
</blockquote></details>

If this is our own field, any reason it is capitalized? SSID is capitalized because it's an acronym.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java, line 71 at r3](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-KgMSKK3u2r6yvNGmpKi:-KgMSKK4Q12W9bCPQMgJ:b7q3mq5) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/ef831968684f98b748627cd0b4707eb0f868efe8/src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java#L71)):*
> ```Java
>         } else if (object instanceof WifiInfo) {
>             return serializeWifiInfo((WifiInfo) object);
>         } else if (object instanceof BluetoothDevice) {
> ```

Mind keeping this if statement alphabetical by class name?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26)*

ygw11223:
Review status: 0 of 2 files reviewed at latest revision, 6 unresolved discussions.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java, line 82 at r1](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-KeWHS6mUdRqNPhOMJSn:-KgrhkXR_5o3tfIycf7_:b1b8uan) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/1bfc5dcf49070354e0fa6e6aaa9d241a985c40f7/src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java#L82)):*
<details><summary><i>Previously, adorokhine (Alexander Dorokhine) wrote…</i></summary><blockquote>

From the same API document, it looks like it persists for about 12 seconds. Given that, do we need to expose it as its own separate RPC? Can we make it a private method and let users use the bluetoothDiscoveryAndGetResults() RPC instead, unless you can think of a use case?
</blockquote></details>

Changed to private.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java, line 118 at r1](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-KeWIoSTrpqtMzytE4xJ:-KgrhvDlY39GJRCwzujx:b-896fix) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/1bfc5dcf49070354e0fa6e6aaa9d241a985c40f7/src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java#L118)):*
<details><summary><i>Previously, adorokhine (Alexander Dorokhine) wrote…</i></summary><blockquote>

Yeah let's make a finally block. bluetoothStartDiscovery() can throw exceptions as well, and if that happens we'll leave the receive registered.

```
mContext.registerReceiver(receiver, filter);
try {
  bluetoothStartDiscovery();
  if (!Utils.waitUntil(() -> mIsScanResultAvailable, 60)) {
    throw new BluetoothAdapterSnippetException(
        "Failed to get discovery results after 1 min, timeout!");
  }
} finally {
  mContext.unregisterReceiver();
}
</blockquote></details>

Done.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java, line 45 at r3](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-KgMRnWD0iiAG8PYLggg:-KgrhdZisVkNXZv_F7e0:b-896fix) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/ef831968684f98b748627cd0b4707eb0f868efe8/src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java#L45)):*
<details><summary><i>Previously, adorokhine (Alexander Dorokhine) wrote…</i></summary><blockquote>

> mDiscoveryResults
Should be `final`.
</blockquote></details>

Done.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java, line 126 at r1](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-KeWJLegiFmNAHv01p1I:-Kgri68hhyiooKDm3ch7:b-896fix) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/1bfc5dcf49070354e0fa6e6aaa9d241a985c40f7/src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java#L126)):*
<details><summary><i>Previously, adorokhine (Alexander Dorokhine) wrote…</i></summary><blockquote>

If this is our own field, any reason it is capitalized? SSID is capitalized because it's an acronym.
</blockquote></details>

Done.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java, line 71 at r3](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-KgMSKK3u2r6yvNGmpKi:-Kgri1v-RiP-64l7MM5i:b-896fix) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/ef831968684f98b748627cd0b4707eb0f868efe8/src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java#L71)):*
<details><summary><i>Previously, adorokhine (Alexander Dorokhine) wrote…</i></summary><blockquote>

Mind keeping this if statement alphabetical by class name?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26)*
<!-- Sent from Reviewable.io -->

**Comment by adorokhine** [comment]
:lgtm:

Looks good to me. Thanks a lot for making this! Leaving final approval to Ang.

---

Reviewed 2 of 2 files at r4.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-:-Kgrkwb_JRDDcb-yM67r:b-z5gp4f)*

**Comment by xpconanfan** [comment]
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java, line 97 at r4](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-KgtxLLz2PClIgePgkQL:-KgtxLLz2PClIgePgkQM:bu1rp2n) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/88e4dce2a4fe3d331bf601fa77a4919c522550d0/src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java#L97)):*
> ```Java
>         description =
>                 "Start discovery, wait for discovery to complete, and return results, which is a list of "
>                         + "serialized BluetoothDevice objects."
> ```

The return values are `JSONObjects`, not actual `BluetoothDevice` objs though?

---

*[src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java, line 103 at r4](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-Kgtxhqxwsrg_9A1AMi6:-Kgtxhqxwsrg_9A1AMi7:br6jrkp) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/88e4dce2a4fe3d331bf601fa77a4919c522550d0/src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java#L103)):*
> ```Java
>         IntentFilter filter = new IntentFilter(BluetoothDevice.ACTION_FOUND);
>         filter.addAction(BluetoothAdapter.ACTION_DISCOVERY_FINISHED);
>         if (mBluetoothAdapter.isDiscovering()) {
> ```

So both this and `bluetoothStartDiscovery` start discovery, but they have inconsistent behavior?
If discovery is already going on, this method will cancel the ongoing discovery and start a new one, whereas `bluetoothStartDiscovery` will silently return and let the ongoing discovery continue.

Perhaps we should be consistent?

---

*[src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java, line 126 at r4](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-Kgtyh2N2UA2DVaIHqVY:-Kgtyh2OKB6ovKxiPsao:byb1sav) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/88e4dce2a4fe3d331bf601fa77a4919c522550d0/src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java#L126)):*
> ```Java
>     private JSONObject serializeBluetoothDevice(BluetoothDevice data) throws JSONException {
>         JSONObject result = new JSONObject(mGson.toJson(data));
>         result.put("Name", data.getName());
> ```

Could we also add the MAC address of the device here?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26)*

**Comment by ygw11223** [comment]
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java, line 97 at r4](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-KgtxLLz2PClIgePgkQL:-KgzBwEcfmK4FttDutcj:bsl912x) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/88e4dce2a4fe3d331bf601fa77a4919c522550d0/src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java#L97)):*

The return values are `JSONObjects`, not actual `BluetoothDevice` objs though?
</blockquote></details>

I referred to the wifi snippet which used similar language.  I think "serialized BluetoothDevice" addressed that the returned JSONObjects are actually BluetoothDevice object.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java, line 103 at r4](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-Kgtxhqxwsrg_9A1AMi6:-KgzCoA-OHkY7PFXIPWM:b-94uhol) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/88e4dce2a4fe3d331bf601fa77a4919c522550d0/src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java#L103)):*

So both this and `bluetoothStartDiscovery` start discovery, but they have inconsistent behavior?
If discovery is already going on, this method will cancel the ongoing discovery and start a new one, whereas `bluetoothStartDiscovery` will silently return and let the ongoing discovery continue.

Perhaps we should be consistent?
</blockquote></details>

I think this method has to kill the existing discovery because the receiver is not registered until we run this method. Since the bluetoothStartDiscovery() has been changed to private and is only used here, maybe I can delete the check of existing discovery there?

---

*[src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java, line 126 at r4](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-Kgtyh2N2UA2DVaIHqVY:-KgzDVz06P_15ALZgYyQ:b9ekql1) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/88e4dce2a4fe3d331bf601fa77a4919c522550d0/src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java#L126)):*

Could we also add the MAC address of the device here?
</blockquote></details>

The mac address is automatically added by the Json builder.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26)*

**Comment by xpconanfan** [comment]
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java, line 78 at r1](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-KeWH3ZflrStro20C4Gi:-KgzEq7KJTlVUvv7_c0q:bcp6nyg) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/1bfc5dcf49070354e0fa6e6aaa9d241a985c40f7/src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java#L78)):*
<details><summary><i>Previously, adorokhine (Alexander Dorokhine) wrote…</i></summary><blockquote>

@Ang Li I thought we weren't doing the namespace prefixes anymore? I'm confused w.r.t our convention.
</blockquote></details>

Discussed offline, let's rename all methods in `BluetoothSnippet` to begin with "bt" instead of "bluetooth".

---

*[src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java, line 103 at r4](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-Kgtxhqxwsrg_9A1AMi6:-KgzF7TmRZ-uL96M4neB:b-5cqi9h) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/88e4dce2a4fe3d331bf601fa77a4919c522550d0/src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java#L103)):*
<details><summary><i>Previously, ygw11223 wrote…</i></summary><blockquote>

I think this method has to kill the existing discovery because the receiver is not registered until we run this method. Since the bluetoothStartDiscovery() has been changed to private and is only used here, maybe I can delete the check of existing discovery there?
</blockquote></details>

Hmm, if `bluetoothStartDiscovery` is a short private method that's only used here, could we merge it into `bluetoothDiscoveryAndGetResults`?

---

*[src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java, line 126 at r4](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-Kgtyh2N2UA2DVaIHqVY:-KgzFLX-mS5c1hi8Aobp:b81wpcd) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/88e4dce2a4fe3d331bf601fa77a4919c522550d0/src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java#L126)):*
<details><summary><i>Previously, ygw11223 wrote…</i></summary><blockquote>

The mac address is automatically added by the Json builder.
</blockquote></details>

Oh, then why do we need to put "Name" separately? Do the other fields have incorrect names?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26)*

**Comment by ygw11223** [comment]
Review status: 1 of 2 files reviewed at latest revision, 3 unresolved discussions.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java, line 78 at r1](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-KeWH3ZflrStro20C4Gi:-Kh0bcLnQAGEi4FrVR_H:b-896fix) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/1bfc5dcf49070354e0fa6e6aaa9d241a985c40f7/src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java#L78)):*

Discussed offline, let's rename all methods in `BluetoothSnippet` to begin with "bt" instead of "bluetooth".
</blockquote></details>

Done.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java, line 103 at r4](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-Kgtxhqxwsrg_9A1AMi6:-Kh0biRd6CdCugwJCeV6:b-896fix) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/88e4dce2a4fe3d331bf601fa77a4919c522550d0/src/main/java/com/google/android/mobly/snippet/bundled/BluetoothAdapterSnippet.java#L103)):*

Hmm, if `bluetoothStartDiscovery` is a short private method that's only used here, could we merge it into `bluetoothDiscoveryAndGetResults`?
</blockquote></details>

Done.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java, line 126 at r4](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-Kgtyh2N2UA2DVaIHqVY:-Kh0brzjTVKxYuXNqL6n:b-m3i4d0) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/88e4dce2a4fe3d331bf601fa77a4919c522550d0/src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java#L126)):*

Oh, then why do we need to put "Name" separately? Do the other fields have incorrect names?
</blockquote></details>

The mac address is the only field that Json builder automatically added. I added "Name" manually because it helps me to distinguish different devices on the python side (name is more readable than mac address). Do we need any other fields?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26)*

**Comment by xpconanfan** [comment]
Review status: 1 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java, line 126 at r4](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-Kgtyh2N2UA2DVaIHqVY:-Kh0nnSSnogmE_Ed0FjZ:bz03924) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/88e4dce2a4fe3d331bf601fa77a4919c522550d0/src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java#L126)):*
<details><summary><i>Previously, ygw11223 wrote…</i></summary><blockquote>

The mac address is the only field that Json builder automatically added. I added "Name" manually because it helps me to distinguish different devices on the python side (name is more readable than mac address). Do we need any other fields?
</blockquote></details>

what's the name of the mac address field in the `JSONObject`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26)*

**Comment by ygw11223** [comment]
Review status: 1 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java, line 126 at r4](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-Kgtyh2N2UA2DVaIHqVY:-Kh5L9jYB-qxnnL-xCEn:b-syd9ro) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/88e4dce2a4fe3d331bf601fa77a4919c522550d0/src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java#L126)):*

what's the name of the mac address field in the `JSONObject`?
</blockquote></details>

I just noticed that it's "mAddress"..Maybe it's better to change "Name" to "mName"?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26)*

**Comment by ygw11223** [comment]
Review status: 1 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java, line 126 at r4](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-Kgtyh2N2UA2DVaIHqVY:-Kh5MG5CdcOt2so7VBvq:bojixa0) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/88e4dce2a4fe3d331bf601fa77a4919c522550d0/src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java#L126)):*
<details><summary><i>Previously, ygw11223 wrote…</i></summary><blockquote>

I just noticed that it's "mAddress"..Maybe it's better to change "Name" to "mName"?
</blockquote></details>

Oh m is mac never mind..

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26)*

**Comment by adorokhine** [comment]
Review status: 1 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java, line 126 at r4](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-Kgtyh2N2UA2DVaIHqVY:-Kh5Mi-v8L5iNNq4MwjJ:b-rzs4wq) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/88e4dce2a4fe3d331bf601fa77a4919c522550d0/src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java#L126)):*
<details><summary><i>Previously, ygw11223 wrote…</i></summary><blockquote>

Oh m is mac never mind..
</blockquote></details>

Are you sure it's mac? I thought gson dumped all instance members by reflection, in which case m would be 'member'.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26)*

**Comment by ygw11223** [comment]
Review status: 1 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java, line 126 at r4](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-Kgtyh2N2UA2DVaIHqVY:-Kh5OjcFcBHeWDgJPlP1:bt1vo98) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/88e4dce2a4fe3d331bf601fa77a4919c522550d0/src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java#L126)):*
<details><summary><i>Previously, adorokhine (Alexander Dorokhine) wrote…</i></summary><blockquote>

Are you sure it's mac? I thought gson dumped all instance members by reflection, in which case m would be 'member'.
</blockquote></details>

I'm not. "Member" makes more sense, and I think I need to manually add the Address field to get rid of this m.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26)*

**Comment by ygw11223** [comment]
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java, line 126 at r4](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-Kgtyh2N2UA2DVaIHqVY:-Kh5uhLFtfIlb5AM72tK:b-bxwv18) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/88e4dce2a4fe3d331bf601fa77a4919c522550d0/src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java#L126)):*
<details><summary><i>Previously, ygw11223 wrote…</i></summary><blockquote>

I'm not. "Member" makes more sense, and I think I need to manually add the Address field to get rid of this m.
</blockquote></details>

Done. Manually added Address as well as other useful info including bond state, type, and supported uuids.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26)*

**Comment by xpconanfan** [comment]
:lgtm:

---

Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/26#-:-Kh8avEa8U4p9qQgg1x7:bnfp4nl)*

File: src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java [hunk 1]
```diff
@@ -117,4 +123,47 @@ private JSONObject serializeWifiInfo(WifiInfo data) throws JSONException {
...
+                result.put("BondState", "None");
+                break;
+            case 11:
+                result.put("BondState", "Bonding");
+                break;
+            case 12:
+                result.put("BondState", "Bonded");
+        }
+        if (data.getName() == null) {
+            result.put("Name", "Null");
```
adorokhine: Why is this null field being transformed into a constant string containing the word "Null"? json can represent real null values.

src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java
```diff
@@ -117,4 +123,47 @@ private JSONObject serializeWifiInfo(WifiInfo data) throws JSONException {
         }
         return result;
     }
+
+    private JSONObject serializeBluetoothDevice(BluetoothDevice data) throws JSONException {
+        JSONObject result = new JSONObject();
+        result.put("Address", data.getAddress());
+        switch (data.getBondState()) {
+            case 10:
```
adorokhine: Where did the magic number 10 come from here? How do you know 10 corresponds to None? How does a future maintainer find out if a new state is needed and what number that state corresponds to?

src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java
```diff
@@ -117,4 +123,47 @@ private JSONObject serializeWifiInfo(WifiInfo data) throws JSONException {
...
+                result.put("Type", "Classic");
+                break;
+            case 2:
+                result.put("Type", "Le");
+                break;
+            case 3:
+                result.put("Type", "Dual");
+        }
+        if (data.getUuids() == null) {
+            result.put("UUIDs", "Null");
```
adorokhine: Inconsistent capitalization versus what the method name is called. Do we have a policy for naming these fields?

src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java
```diff
@@ -117,4 +123,47 @@ private JSONObject serializeWifiInfo(WifiInfo data) throws JSONException {
...
+                result.put("BondState", "Bonded");
+        }
+        if (data.getName() == null) {
+            result.put("Name", "Null");
+        } else {
+            result.put("Name", data.getName());
+        }
+        switch (data.getType()) {
+            case 0:
+                result.put("Type", "Unknown");
```
adorokhine: Do the other json results we serialize use the same capitalization? Could we have examples for serialized versions of all the types we support in the javadoc comments of this class?

src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java
```diff
@@ -117,4 +123,47 @@ private JSONObject serializeWifiInfo(WifiInfo data) throws JSONException {
...
+                break;
+            case 1:
```
adorokhine: Same magic number concern as above.

PR #42, opened 2017-05-02
Throw exceptions in case of failure to enable or disable HCI snoop log.

adorokhine:
Returning boolean is not consistent with our style and is not done for any
other method in MBS.

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly-bundled-snippets/42)


xpconanfan:
Is there a better way to do reflection so we don't have to repeat this similar try/catch block every time?

adorokhine:
Sure, let me take a stab at it in a different PR.

PR #80, opened 2017-09-26
Method to perform HTTP download to local storage.

kdart:
Use Android API to download file using HTTP and store locally. Locally compute MD5 hash and  respond to host with Uri and hash value.

Since current usage of curl does not make use of the downloaded file other than to compute the MD5 hash to, presumably, verify transfer integrity it's not necessary to copy the file to the host.

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly-bundled-snippets/80)

### DIFF
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -39,6 +39,7 @@
                            com.google.android.mobly.snippet.bundled.NotificationSnippet,
                            com.google.android.mobly.snippet.bundled.TelephonySnippet,
                            com.google.android.mobly.snippet.bundled.NetworkingSnippet,
+                           com.google.android.mobly.snippet.bundled.FileSnippet,
                            com.google.android.mobly.snippet.bundled.SmsSnippet,
                            com.google.android.mobly.snippet.bundled.WifiManagerSnippet" />
     </application>
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -13,6 +13,7 @@
     <uses-permission android:name="android.permission.CHANGE_WIFI_STATE" />
     <uses-permission android:name="android.permission.GET_ACCOUNTS" />
     <uses-permission android:name="android.permission.INTERNET" />
+    <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
     <uses-permission android:name="android.permission.MANAGE_ACCOUNTS" />
     <uses-permission android:name="android.permission.MODIFY_AUDIO_SETTINGS" />
     <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE" />
--- a/src/main/java/com/google/android/mobly/snippet/bundled/FileSnippet.java
+++ b/src/main/java/com/google/android/mobly/snippet/bundled/FileSnippet.java
@@ -64,11 +64,10 @@ public class FileSnippet implements Snippet {
         }
     }
 
-    @Rpc(description = "Remove a file pointed to by the content URI. Return "
-            + "number deleted (should always be one)")
-    public int fileDeleteContent(String uri) {
+    @Rpc(description = "Remove a file pointed to by the content URI.")
+    public void fileDeleteContent(String uri) {
         Uri uri_ = Uri.parse(uri);
-        return mContext.getContentResolver().delete(uri_, null, null);
+        mContext.getContentResolver().delete(uri_, null, null);
     }
 
     @Override
--- a/src/main/java/com/google/android/mobly/snippet/bundled/FileSnippet.java
+++ b/src/main/java/com/google/android/mobly/snippet/bundled/FileSnippet.java
@@ -27,7 +27,6 @@ import android.support.test.InstrumentationRegistry;
 import com.google.android.mobly.snippet.Snippet;
 import com.google.android.mobly.snippet.bundled.utils.Utils;
 import com.google.android.mobly.snippet.rpc.Rpc;
-import com.google.android.mobly.snippet.util.Log;
 
 /** Snippet class for File and abstract storage URI operation RPCs. */
 public class FileSnippet implements Snippet {
--- a/src/main/java/com/google/android/mobly/snippet/bundled/FileSnippet.java
+++ b/src/main/java/com/google/android/mobly/snippet/bundled/FileSnippet.java
@@ -1,0 +1,70 @@
+/*
+ * Copyright (C) 2017 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.android.mobly.snippet.bundled;
+
+import java.io.IOException;
+import java.security.MessageDigest;
+import java.security.DigestInputStream;
+import java.security.NoSuchAlgorithmException;
+import android.content.Context;
+import android.net.Uri;
+import android.os.ParcelFileDescriptor;
+import android.support.test.InstrumentationRegistry;
+import com.google.android.mobly.snippet.Snippet;
+import com.google.android.mobly.snippet.bundled.utils.Utils;
+import com.google.android.mobly.snippet.rpc.Rpc;
+import com.google.android.mobly.snippet.util.Log;
+
+/** Snippet class for File and abstract storage URI operation RPCs. */
+public class FileSnippet implements Snippet {
+
+    private final Context mContext;
+
+    public FileSnippet() {
+        mContext = InstrumentationRegistry.getContext();
+    }
+
+    private static class FileSnippetException extends Exception {
+
+        private static final long serialVersionUID = 8081L;
+
+        public FileSnippetException(String msg) {
+            super(msg);
+        }
+    }
+
+    @Rpc(description = "Compute MD5 hash on a content URI. Return the MD5 has has a hex string.")
+    public String fileMd5Hash(String uri) throws IOException, NoSuchAlgorithmException  {
+        Uri uri_ = Uri.parse(uri);
+        ParcelFileDescriptor pfd = mContext.getContentResolver().openFileDescriptor(uri_, "r");
+        MessageDigest md = MessageDigest.getInstance("MD5");
+        int length = (int) pfd.getStatSize();
+        byte[] buf = new byte[length];
+        ParcelFileDescriptor.AutoCloseInputStream stream = new ParcelFileDescriptor.AutoCloseInputStream(pfd);
+        DigestInputStream dis = new DigestInputStream(stream, md);
+        try {
+            dis.read(buf, 0, length);
+            return Utils.bytesToHexString(md.digest());
+        } finally {
+            dis.close();
+            stream.close();
+        }
+    }
+
+    @Override
+    public void shutdown() {}
+}
--- a/src/main/java/com/google/android/mobly/snippet/bundled/FileSnippet.java
+++ b/src/main/java/com/google/android/mobly/snippet/bundled/FileSnippet.java
@@ -64,6 +64,13 @@ public class FileSnippet implements Snippet {
         }
     }
 
+    @Rpc(description = "Remove a file pointed to by the content URI. Return "
+            + "number deleted (should always be one)")
+    public int fileDeleteContent(String uri) {
+        Uri uri_ = Uri.parse(uri);
+        return mContext.getContentResolver().delete(uri_, null, null);
+    }
+
     @Override
     public void shutdown() {}
 }
--- a/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java
+++ b/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java
@@ -20,12 +20,30 @@ import java.net.InetAddress;
 import java.net.Socket;
 import java.io.IOException;
 import java.net.UnknownHostException;
+import android.content.Context;
+import android.content.Intent;
+import android.content.IntentFilter;
+import android.content.BroadcastReceiver;
+import android.net.Uri;
+import android.os.Environment;
+import android.app.DownloadManager;
+import android.support.test.InstrumentationRegistry;
 import com.google.android.mobly.snippet.Snippet;
+import com.google.android.mobly.snippet.bundled.utils.Utils;
 import com.google.android.mobly.snippet.rpc.Rpc;
 import com.google.android.mobly.snippet.util.Log;
 
 /** Snippet class for networking RPCs. */
 public class NetworkingSnippet implements Snippet {
+
+    private final DownloadManager mDownloadManager;
+    private final Context mContext;
+    private volatile boolean mIsDownloadComplete = false;
+
+    public NetworkingSnippet() {
+        mContext = InstrumentationRegistry.getContext();
+        mDownloadManager = (DownloadManager) mContext.getSystemService(Context.DOWNLOAD_SERVICE);
+    }
 
     @Rpc(description = "Check if a host and port are connectable using a TCP connection attempt.")
     public boolean networkIsTcpConnectable(String host, int port) {
@@ -45,6 +63,46 @@ public class NetworkingSnippet implements Snippet {
             return false;
         }
         return true;
+    }
+
+    @Rpc(description = "Download a file using HTTP.")
+    public String networkHTTPDownload(String url, String destination) {
+        long reqid = 0;
+        Uri uri = Uri.parse(url);
+        DownloadManager.Request request = new DownloadManager.Request(uri);
+        request.setDestinationInExternalPublicDir(Environment.DIRECTORY_DOWNLOADS, destination);
+        mIsDownloadComplete = false;
+        IntentFilter filter = new IntentFilter(DownloadManager.ACTION_DOWNLOAD_COMPLETE);
+        BroadcastReceiver receiver = new DownloadReceiver();
+        mContext.registerReceiver(receiver, filter);
+        try {
+            reqid = mDownloadManager.enqueue(request);
+            Log.d(String.format("networkHTTPDownload download of %s with id %d", url, reqid));
+            if (!Utils.waitUntil(() -> mIsDownloadComplete, 240)) {
+                return "";
+            }
+        } finally {
+            mContext.unregisterReceiver(receiver);
+        }
+
+        Uri resp = mDownloadManager.getUriForDownloadedFile(reqid);
+        if (resp != null) {
+            Log.d(String.format("networkHTTPDownload completed to %s", resp.toString()));
+            return resp.toString();
+        } else {
+            return "";
+        }
+    }
+
+    private class DownloadReceiver extends BroadcastReceiver {
+
+        @Override
+        public void onReceive(Context context, Intent intent) {
+            String action = intent.getAction();
+            if (DownloadManager.ACTION_DOWNLOAD_COMPLETE.equals(action)) {
+                mIsDownloadComplete = true;
+            }
+        }
     }
 
     @Override
--- a/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java
+++ b/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java
@@ -107,6 +107,7 @@ public class NetworkingSnippet implements Snippet {
         Uri resp = mDownloadManager.getUriForDownloadedFile(mReqid);
         if (resp != null) {
             Log.d(String.format("networkHttpDownload completed to %s", resp.toString()));
+            mReqid = 0;
             return resp.toString();
         } else {
             Log.d(String.format("networkHttpDownload Failed to download %s", uri.toString()));
@@ -128,5 +129,9 @@ public class NetworkingSnippet implements Snippet {
     }
 
     @Override
-    public void shutdown() {}
+    public void shutdown() {
+        if (mReqid != 0) {
+            mDownloadManager.remove(mReqid);
+        }
+    }
 }
--- a/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java
+++ b/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java
@@ -16,9 +16,7 @@
 
 package com.google.android.mobly.snippet.bundled;
 
-import java.util.Map;
 import java.util.List;
-import java.util.HashMap;
 import java.net.InetAddress;
 import java.net.Socket;
 import java.io.IOException;
@@ -26,8 +24,8 @@ import java.net.UnknownHostException;
 import java.security.MessageDigest;
 import java.security.DigestInputStream;
 import java.security.NoSuchAlgorithmException;
-import android.content.Context;
 import android.content.Intent;
+import android.content.Context;
 import android.content.IntentFilter;
 import android.content.BroadcastReceiver;
 import android.net.Uri;
@@ -43,13 +41,22 @@ import com.google.android.mobly.snippet.util.Log;
 /** Snippet class for networking RPCs. */
 public class NetworkingSnippet implements Snippet {
 
-    private final DownloadManager mDownloadManager;
     private final Context mContext;
+    private final DownloadManager mDownloadManager;
     private volatile boolean mIsDownloadComplete = false;
 
     public NetworkingSnippet() {
         mContext = InstrumentationRegistry.getContext();
         mDownloadManager = (DownloadManager) mContext.getSystemService(Context.DOWNLOAD_SERVICE);
+    }
+
+    private static class NetworkingSnippetException extends Exception {
+
+        private static final long serialVersionUID = 8080L;
+
+        public NetworkingSnippetException(String msg) {
+            super(msg);
+        }
     }
 
     @Rpc(description = "Check if a host and port are connectable using a TCP connection attempt.")
@@ -73,16 +80,15 @@ public class NetworkingSnippet implements Snippet {
     }
 
     @Rpc(description = "Download a file using HTTP. Return content Uri and MD5 hash value.")
-    public Map<String, String> networkHTTPDownload(String url) throws IOException {
+    public String networkHTTPDownload(String url) throws IllegalArgumentException, NetworkingSnippetException {
         long reqid = 0;
-        MessageDigest md;
-        ParcelFileDescriptor pfd;
 
         Uri uri = Uri.parse(url);
         List<String> pathsegments = uri.getPathSegments();
         if (pathsegments.size() < 1) {
-            Log.d(String.format("The Uri %s does not have a path.", uri.toString()));
-            return null;
+            String msg = String.format("The Uri %s does not have a path.", uri.toString());
+            Log.d(msg);
+            throw new IllegalArgumentException(msg);
         }
         DownloadManager.Request request = new DownloadManager.Request(uri);
         request.setDestinationInExternalPublicDir(Environment.DIRECTORY_DOWNLOADS,
@@ -96,39 +102,42 @@ public class NetworkingSnippet implements Snippet {
             Log.d(String.format("networkHTTPDownload download of %s with id %d", url, reqid));
             if (!Utils.waitUntil(() -> mIsDownloadComplete, 30)) {
                 Log.d(String.format("networkHTTPDownload timed out waiting for completion"));
-                return null;
+                throw new NetworkingSnippetException("networkHTTPDownload timed out.");
             }
         } finally {
             mContext.unregisterReceiver(receiver);
         }
-        // Now compute the MD5 hash of downloaded file and return it with the
-        // destination Uri (which is an abstract content Uri).
         Uri resp = mDownloadManager.getUriForDownloadedFile(reqid);
         if (resp != null) {
             Log.d(String.format("networkHTTPDownload completed to %s", resp.toString()));
+            return resp.toString();
+        } else {
+            Log.d(String.format("networkHTTPDownload Failed to download %s", uri.toString()));
+            throw new NetworkingSnippetException("networkHTTPDownload didn't get downloaded file.");
+        }
+    }
 
-            try {
-                md = MessageDigest.getInstance("MD5");
-            // This should never happen, but we have to make Java happy.
-            } catch (NoSuchAlgorithmException algerr) {
-                return null;
-            }
-            pfd = mDownloadManager.openDownloadedFile(reqid);
-            ParcelFileDescriptor.AutoCloseInputStream stream = new ParcelFileDescriptor.AutoCloseInputStream(pfd);
-            int length = (int) pfd.getStatSize();
-            DigestInputStream dis = new DigestInputStream(stream, md);
-            byte[] buf = new byte[length];
+    @Rpc(description = "Compute MD5 hash on a content URI.")
+    public String networkMD5Hash(String url) throws IOException, NoSuchAlgorithmException  {
+        ParcelFileDescriptor pfd;
+        MessageDigest md;
+        String hexdigest = "";
+
+        Uri uri = Uri.parse(url);
+        pfd = mContext.getContentResolver().openFileDescriptor(uri, "r");
+        md = MessageDigest.getInstance("MD5");
+        int length = (int) pfd.getStatSize();
+        byte[] buf = new byte[length];
+        ParcelFileDescriptor.AutoCloseInputStream stream = new ParcelFileDescriptor.AutoCloseInputStream(pfd);
+        DigestInputStream dis = new DigestInputStream(stream, md);
+        try {
             dis.read(buf, 0, length);
-            String hexdigest = Utils.bytesToHex(md.digest());
+            hexdigest = Utils.bytesToHexString(md.digest());
+        } finally {
             dis.close();
             stream.close();
-            Map<String, String> rv = new HashMap<String, String>();
-            rv.put("Uri", resp.toString());
-            rv.put("md5", hexdigest);
-            return rv;
-        } else {
-            return null;
         }
+        return hexdigest;
     }
 
     private class DownloadReceiver extends BroadcastReceiver {
--- a/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java
+++ b/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java
@@ -16,11 +16,12 @@
 
 package com.google.android.mobly.snippet.bundled;
 
-import java.util.*;
+import java.util.Map;
+import java.util.List;
+import java.util.HashMap;
 import java.net.InetAddress;
 import java.net.Socket;
 import java.io.IOException;
-import java.io.FileNotFoundException;
 import java.net.UnknownHostException;
 import java.security.MessageDigest;
 import java.security.DigestInputStream;
@@ -44,7 +45,6 @@ public class NetworkingSnippet implements Snippet {
 
     private final DownloadManager mDownloadManager;
     private final Context mContext;
-    private final static char[] hexArray = "0123456789abcdef".toCharArray();
     private volatile boolean mIsDownloadComplete = false;
 
     public NetworkingSnippet() {
@@ -73,14 +73,20 @@ public class NetworkingSnippet implements Snippet {
     }
 
     @Rpc(description = "Download a file using HTTP. Return content Uri and MD5 hash value.")
-    public Map<String, String> networkHTTPDownload(String url, String destination) throws IOException {
+    public Map<String, String> networkHTTPDownload(String url) throws IOException {
         long reqid = 0;
         MessageDigest md;
         ParcelFileDescriptor pfd;
 
         Uri uri = Uri.parse(url);
+        List<String> pathsegments = uri.getPathSegments();
+        if (pathsegments.size() < 1) {
+            Log.d(String.format("The Uri %s does not have a path.", uri.toString()));
+            return null;
+        }
         DownloadManager.Request request = new DownloadManager.Request(uri);
-        request.setDestinationInExternalPublicDir(Environment.DIRECTORY_DOWNLOADS, destination);
+        request.setDestinationInExternalPublicDir(Environment.DIRECTORY_DOWNLOADS,
+                                                  pathsegments.get(pathsegments.size() - 1));
         mIsDownloadComplete = false;
         IntentFilter filter = new IntentFilter(DownloadManager.ACTION_DOWNLOAD_COMPLETE);
         BroadcastReceiver receiver = new DownloadReceiver();
@@ -88,14 +94,15 @@ public class NetworkingSnippet implements Snippet {
         try {
             reqid = mDownloadManager.enqueue(request);
             Log.d(String.format("networkHTTPDownload download of %s with id %d", url, reqid));
-            if (!Utils.waitUntil(() -> mIsDownloadComplete, 240)) {
+            if (!Utils.waitUntil(() -> mIsDownloadComplete, 30)) {
+                Log.d(String.format("networkHTTPDownload timed out waiting for completion"));
                 return null;
             }
         } finally {
             mContext.unregisterReceiver(receiver);
         }
-
-        // Now compute the MD5 hash of downloaded file and return it.
+        // Now compute the MD5 hash of downloaded file and return it with the
+        // destination Uri (which is an abstract content Uri).
         Uri resp = mDownloadManager.getUriForDownloadedFile(reqid);
         if (resp != null) {
             Log.d(String.format("networkHTTPDownload completed to %s", resp.toString()));
@@ -112,7 +119,7 @@ public class NetworkingSnippet implements Snippet {
             DigestInputStream dis = new DigestInputStream(stream, md);
             byte[] buf = new byte[length];
             dis.read(buf, 0, length);
-            String hexdigest = bytesToHex(md.digest());
+            String hexdigest = Utils.bytesToHex(md.digest());
             dis.close();
             stream.close();
             Map<String, String> rv = new HashMap<String, String>();
@@ -134,17 +141,6 @@ public class NetworkingSnippet implements Snippet {
             }
         }
     }
-
-
-    public static String bytesToHex(byte[] bytes) {
-        char[] hexChars = new char[bytes.length * 2];
-        for ( int j = 0; j < bytes.length; j++ ) {
-            int v = bytes[j] & 0xFF;
-            hexChars[j * 2] = hexArray[v >>> 4];
-            hexChars[j * 2 + 1] = hexArray[v & 0x0F];
-        }
-        return new String(hexChars);
-}
 
     @Override
     public void shutdown() {}
--- a/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java
+++ b/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java
@@ -16,6 +16,7 @@
 
 package com.google.android.mobly.snippet.bundled;
 
+import java.util.*;
 import java.net.InetAddress;
 import java.net.Socket;
 import java.io.IOException;
@@ -71,8 +72,8 @@ public class NetworkingSnippet implements Snippet {
         return true;
     }
 
-    @Rpc(description = "Download a file using HTTP.")
-    public String networkHTTPDownload(String url, String destination) throws IOException {
+    @Rpc(description = "Download a file using HTTP. Return content Uri and MD5 hash value.")
+    public Map<String, String> networkHTTPDownload(String url, String destination) throws IOException {
         long reqid = 0;
         MessageDigest md;
         ParcelFileDescriptor pfd;
@@ -88,7 +89,7 @@ public class NetworkingSnippet implements Snippet {
             reqid = mDownloadManager.enqueue(request);
             Log.d(String.format("networkHTTPDownload download of %s with id %d", url, reqid));
             if (!Utils.waitUntil(() -> mIsDownloadComplete, 240)) {
-                return "";
+                return null;
             }
         } finally {
             mContext.unregisterReceiver(receiver);
@@ -103,7 +104,7 @@ public class NetworkingSnippet implements Snippet {
                 md = MessageDigest.getInstance("MD5");
             // This should never happen, but we have to make Java happy.
             } catch (NoSuchAlgorithmException algerr) {
-                return "";
+                return null;
             }
             pfd = mDownloadManager.openDownloadedFile(reqid);
             ParcelFileDescriptor.AutoCloseInputStream stream = new ParcelFileDescriptor.AutoCloseInputStream(pfd);
@@ -114,14 +115,14 @@ public class NetworkingSnippet implements Snippet {
             String hexdigest = bytesToHex(md.digest());
             dis.close();
             stream.close();
-            // return resp.toString(); TODO(dart) return the tuple of resp Uri
-            // and hexdigest.
-            return hexdigest;
+            Map<String, String> rv = new HashMap<String, String>();
+            rv.put("Uri", resp.toString());
+            rv.put("md5", hexdigest);
+            return rv;
         } else {
-            return "";
+            return null;
         }
     }
-
 
     private class DownloadReceiver extends BroadcastReceiver {
 
--- a/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java
+++ b/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java
@@ -44,6 +44,7 @@ public class NetworkingSnippet implements Snippet {
     private final Context mContext;
     private final DownloadManager mDownloadManager;
     private volatile boolean mIsDownloadComplete = false;
+    private volatile long mReqid = 0;
 
     public NetworkingSnippet() {
         mContext = InstrumentationRegistry.getContext();
@@ -79,27 +80,26 @@ public class NetworkingSnippet implements Snippet {
         return true;
     }
 
-    @Rpc(description = "Download a file using HTTP. Return content Uri and MD5 hash value.")
+    @Rpc(description = "Download a file using HTTP. Return content Uri (file remains on device). "
+                       + "The Uri should be treated as an opaque handle for further operations.")
     public String networkHTTPDownload(String url) throws IllegalArgumentException, NetworkingSnippetException {
-        long reqid = 0;
 
         Uri uri = Uri.parse(url);
         List<String> pathsegments = uri.getPathSegments();
         if (pathsegments.size() < 1) {
-            String msg = String.format("The Uri %s does not have a path.", uri.toString());
-            Log.d(msg);
-            throw new IllegalArgumentException(msg);
+            throw new IllegalArgumentException(String.format("The Uri %s does not have a path.", uri.toString()));
         }
         DownloadManager.Request request = new DownloadManager.Request(uri);
         request.setDestinationInExternalPublicDir(Environment.DIRECTORY_DOWNLOADS,
                                                   pathsegments.get(pathsegments.size() - 1));
         mIsDownloadComplete = false;
+        mReqid = 0;
         IntentFilter filter = new IntentFilter(DownloadManager.ACTION_DOWNLOAD_COMPLETE);
         BroadcastReceiver receiver = new DownloadReceiver();
         mContext.registerReceiver(receiver, filter);
         try {
-            reqid = mDownloadManager.enqueue(request);
-            Log.d(String.format("networkHTTPDownload download of %s with id %d", url, reqid));
+            mReqid = mDownloadManager.enqueue(request);
+            Log.d(String.format("networkHTTPDownload download of %s with id %d", url, mReqid));
             if (!Utils.waitUntil(() -> mIsDownloadComplete, 30)) {
                 Log.d(String.format("networkHTTPDownload timed out waiting for completion"));
                 throw new NetworkingSnippetException("networkHTTPDownload timed out.");
@@ -107,7 +107,7 @@ public class NetworkingSnippet implements Snippet {
         } finally {
             mContext.unregisterReceiver(receiver);
         }
-        Uri resp = mDownloadManager.getUriForDownloadedFile(reqid);
+        Uri resp = mDownloadManager.getUriForDownloadedFile(mReqid);
         if (resp != null) {
             Log.d(String.format("networkHTTPDownload completed to %s", resp.toString()));
             return resp.toString();
@@ -117,27 +117,25 @@ public class NetworkingSnippet implements Snippet {
         }
     }
 
-    @Rpc(description = "Compute MD5 hash on a content URI.")
-    public String networkMD5Hash(String url) throws IOException, NoSuchAlgorithmException  {
-        ParcelFileDescriptor pfd;
-        MessageDigest md;
-        String hexdigest = "";
+    @Rpc(description = "Compute MD5 hash on a content URI. Return the MD5 has has a hex string.")
+    public String networkMD5Hash(String uri) throws IOException, NoSuchAlgorithmException  {
+        String md5string = "";
 
-        Uri uri = Uri.parse(url);
-        pfd = mContext.getContentResolver().openFileDescriptor(uri, "r");
-        md = MessageDigest.getInstance("MD5");
+        Uri uri_ = Uri.parse(uri);
+        ParcelFileDescriptor pfd = mContext.getContentResolver().openFileDescriptor(uri_, "r");
+        MessageDigest md = MessageDigest.getInstance("MD5");
         int length = (int) pfd.getStatSize();
         byte[] buf = new byte[length];
         ParcelFileDescriptor.AutoCloseInputStream stream = new ParcelFileDescriptor.AutoCloseInputStream(pfd);
         DigestInputStream dis = new DigestInputStream(stream, md);
         try {
             dis.read(buf, 0, length);
-            hexdigest = Utils.bytesToHexString(md.digest());
+            md5string = Utils.bytesToHexString(md.digest());
         } finally {
             dis.close();
             stream.close();
         }
-        return hexdigest;
+        return md5string;
     }
 
     private class DownloadReceiver extends BroadcastReceiver {
@@ -145,7 +143,9 @@ public class NetworkingSnippet implements Snippet {
         @Override
         public void onReceive(Context context, Intent intent) {
             String action = intent.getAction();
-            if (DownloadManager.ACTION_DOWNLOAD_COMPLETE.equals(action)) {
+            long gotid = (long) intent.getExtras().get("extra_download_id");
+            if (DownloadManager.ACTION_DOWNLOAD_COMPLETE.equals(action)
+                            && gotid == mReqid) {
                 mIsDownloadComplete = true;
             }
         }
--- a/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java
+++ b/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java
@@ -21,9 +21,6 @@ import java.net.InetAddress;
 import java.net.Socket;
 import java.io.IOException;
 import java.net.UnknownHostException;
-import java.security.MessageDigest;
-import java.security.DigestInputStream;
-import java.security.NoSuchAlgorithmException;
 import android.content.Intent;
 import android.content.Context;
 import android.content.IntentFilter;
@@ -82,7 +79,7 @@ public class NetworkingSnippet implements Snippet {
 
     @Rpc(description = "Download a file using HTTP. Return content Uri (file remains on device). "
                        + "The Uri should be treated as an opaque handle for further operations.")
-    public String networkHTTPDownload(String url) throws IllegalArgumentException, NetworkingSnippetException {
+    public String networkHttpDownload(String url) throws IllegalArgumentException, NetworkingSnippetException {
 
         Uri uri = Uri.parse(url);
         List<String> pathsegments = uri.getPathSegments();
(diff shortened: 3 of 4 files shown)


kdart:
Usage looks like this.

> ⇒ snippet_shell --serial HT73Y0201123 --mbs
> \>\>\> s.networkHTTPDownload("http://www.google.com/images/branding/googleg/1x/googleg_standard_color_128dp.png", "googleg_standard_color_128dp.png")
> {'Uri': 'content://downloads/all_downloads/29', 'md5': '116d7abaea20868404cb366b611c69b7'}

This might not be the best API for this, since the target name isn't really seen.

xpconanfan:
Review status: 0 of 2 files reviewed at latest revision, 5 unresolved discussions.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 19 at r2](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-KuzOpuTYnOlmyumgkeF:-KuzOpuTYnOlmyumgkeG:b-e6gcfh) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/ed10e12060de46fdec39643f90311779e80d22b8/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L19)):*
> ```Java
> package com.google.android.mobly.snippet.bundled;
> 
> import java.util.*;
> ```

import *?
let's not...

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 76 at r2](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-KuzOwA_e1rdf-n5Xi6U:-KuzOwA_e1rdf-n5Xi6V:b-61mci2) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/ed10e12060de46fdec39643f90311779e80d22b8/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L76)):*
> ```Java
> 
>     @Rpc(description = "Download a file using HTTP. Return content Uri and MD5 hash value.")
>     public Map<String, String> networkHTTPDownload(String url, String destination) throws IOException {
> ```

what does `destination` entail here?
Android file management is brittle and we should be more specific about what `destination` means.
Let's add a docstring with more context.

Also, what if the destination is not writable?

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 98 at r2](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-KuzNxx7gQ0E-c-U0RcA:-KuzNxx8vT-LQQD9gdky:beq3mg6) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/ed10e12060de46fdec39643f90311779e80d22b8/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L98)):*
> ```Java
>         }
> 
>         // Now compute the MD5 hash of downloaded file and return it.
> ```

can the MD5 calculation be a separate Rpc that takes a Uri?

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 132 at r2](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-KuzPGz_30TLm2csrY0P:-KuzPGz_30TLm2csrY0Q:b-pd2kye) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/ed10e12060de46fdec39643f90311779e80d22b8/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L132)):*
> ```Java
>         public void onReceive(Context context, Intent intent) {
>             String action = intent.getAction();
>             if (DownloadManager.ACTION_DOWNLOAD_COMPLETE.equals(action)) {
> ```

should we check that the intent is actually for the file the Rpc wants to download?

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 139 at r2](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-KuzPMI3vnYVahPKbBvY:-KuzPMI3vnYVahPKbBvZ:b-d2jcab) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/ed10e12060de46fdec39643f90311779e80d22b8/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L139)):*
> ```Java
> 
> 
>     public static String bytesToHex(byte[] bytes) {
> ```

`private`?
Or move it to a different util module?

Also do we really have to write code for this? is there no existing API to call?

Need docstring too.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80)*

kdart:
Review status: 0 of 2 files reviewed at latest revision, 5 unresolved discussions.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 19 at r2](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-KuzOpuTYnOlmyumgkeF:-Kv-0-PGSl7rPE9WLIUT:b-896fix) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/ed10e12060de46fdec39643f90311779e80d22b8/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L19)):*

import *?
let's not...
</blockquote></details>

Done.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 76 at r2](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-KuzOwA_e1rdf-n5Xi6U:-KuzyzcJKxRRUDtSofON:bcris2w) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/ed10e12060de46fdec39643f90311779e80d22b8/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L76)):*

what does `destination` entail here?
Android file management is brittle and we should be more specific about what `destination` means.
Let's add a docstring with more context.

Also, what if the destination is not writable?
</blockquote></details>

I didn't really like that, either. I removed it. It needed to be just a base file name. Now I just use the last part of the Uri for the destination name.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 98 at r2](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-KuzNxx7gQ0E-c-U0RcA:-KuzzLYOKsc9_Oywg3Qw:bkjowjh) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/ed10e12060de46fdec39643f90311779e80d22b8/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L98)):*

can the MD5 calculation be a separate Rpc that takes a Uri?
</blockquote></details>

It could be. But I wanted to do it there because  1) in the current use case it's needed immediately, and 2) it then needs only one RPC round-trip instead of two. That halves the odds of getting one of those RPC errors...

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 132 at r2](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-KuzPGz_30TLm2csrY0P:-Kv-2u_3uGZGdLh50Ezx:b36xt79) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/ed10e12060de46fdec39643f90311779e80d22b8/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L132)):*

should we check that the intent is actually for the file the Rpc wants to download?
</blockquote></details>

Filtering API doc is unclear (I couldn't get it to work) and the intent data is null. So I'm not sure how to do that. It's unlikely that it will be something different during this time span. 

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 139 at r2](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-KuzPMI3vnYVahPKbBvY:-Kv--M2CE3mUL9rHtaZ0:b-z2imnq) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/ed10e12060de46fdec39643f90311779e80d22b8/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L139)):*

`private`?
Or move it to a different util module?

Also do we really have to write code for this? is there no existing API to call?

Need docstring too.
</blockquote></details>

Moved to Utils. I looked for something similar but couldn't find anything suitable. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80)*

kdart:
News usage:

> s.networkHTTPDownload("http://www.google.com/images/branding/googleg/1x/googleg_standard_color_128dp.png")
> {'Uri': 'content://downloads/all_downloads/75', 'md5': '116d7abaea20868404cb366b611c69b7'}

---

Review status: 0 of 3 files reviewed at latest revision, 5 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-:-Kv-FY5f0IgEa-ULD4K_:b2avy1z)*

xpconanfan:
Review status: 0 of 3 files reviewed at latest revision, 4 unresolved discussions.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 76 at r2](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-KuzOwA_e1rdf-n5Xi6U:-Kv-O--LfABTnYcL34Fa:b-r2oylf) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/ed10e12060de46fdec39643f90311779e80d22b8/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L76)):*
<details><summary><i>Previously, kdart (Keith Dart) wrote…</i></summary><blockquote>

I didn't really like that, either. I removed it. It needed to be just a base file name. Now I just use the last part of the Uri for the destination name.
</blockquote></details>

It is nice to be able to specify the location to use though.
E.g. if a phone has SD card, we would prefer to use SD card.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 98 at r2](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-KuzNxx7gQ0E-c-U0RcA:-Kv-OA8bX5gxUEzYEZfs:b-50lhyd) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/ed10e12060de46fdec39643f90311779e80d22b8/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L98)):*
<details><summary><i>Previously, kdart (Keith Dart) wrote…</i></summary><blockquote>

It could be. But I wanted to do it there because  1) in the current use case it's needed immediately, and 2) it then needs only one RPC round-trip instead of two. That halves the odds of getting one of those RPC errors...
</blockquote></details>

Seems odd to have extra feature in the download Rpc.
If we add a download from ftp rpc, do we want to do md5 as well, what about other downloads?
We are setting a convention here and let's keep these Rpcs small in scope.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 132 at r2](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-KuzPGz_30TLm2csrY0P:-Kv-OXmoAq525Mg2Kvvf:b-o9x49b) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/ed10e12060de46fdec39643f90311779e80d22b8/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L132)):*
<details><summary><i>Previously, kdart (Keith Dart) wrote…</i></summary><blockquote>

Filtering API doc is unclear (I couldn't get it to work) and the intent data is null. So I'm not sure how to do that. It's unlikely that it will be something different during this time span. 
</blockquote></details>

i wouldn't bet on it.
any app could download anything at any time during the test on a device.
you don't want this to be mistakenly flipped, which will make the test flaky.

Let's figure out a way.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80)*

kdart:
Review status: 0 of 3 files reviewed at latest revision, 4 unresolved discussions.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 76 at r2](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-KuzOwA_e1rdf-n5Xi6U:-Kv-PvkCS1wON_0ZFeyl:b-2qvxw0) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/ed10e12060de46fdec39643f90311779e80d22b8/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L76)):*

It is nice to be able to specify the location to use though.
E.g. if a phone has SD card, we would prefer to use SD card.
</blockquote></details>

Well, this API is very abstract. It takes the given Uri and downloads to some place we don't know the details of. If you use the method setDestinationInExternalPublicDir, as is done here, that basically means /sdcard (or it's emulation). So that should already be the case. 

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 98 at r2](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-KuzNxx7gQ0E-c-U0RcA:-Kv-QQ7VI0KQr_fDkPHy:bbducw5) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/ed10e12060de46fdec39643f90311779e80d22b8/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L98)):*

Seems odd to have extra feature in the download Rpc.
If we add a download from ftp rpc, do we want to do md5 as well, what about other downloads?
We are setting a convention here and let's keep these Rpcs small in scope.
</blockquote></details>

Yes, I think you are right.  "Damn the RPC Errors, full speed ahead!" 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80)*

xpconanfan:
Review status: 0 of 3 files reviewed at latest revision, 5 unresolved discussions.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 84 at r3](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-Kv-asBcqwFKT3J8DZjo:-Kv-asBcqwFKT3J8DZjp:b-90xxor) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/0c43c4fa602b08a96bfc54a541cd48c7df111aeb/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L84)):*
> ```Java
>         List<String> pathsegments = uri.getPathSegments();
>         if (pathsegments.size() < 1) {
>             Log.d(String.format("The Uri %s does not have a path.", uri.toString()));
> ```

we should throw exception instead of returning null

---

*[src/main/java/com/google/android/mobly/snippet/bundled/utils/Utils.java, line 205 at r3](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-Kv-bGEZxGip9vyOMZ8O:-Kv-bGEZxGip9vyOMZ8P:b7cyknz) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/0c43c4fa602b08a96bfc54a541cd48c7df111aeb/src/main/java/com/google/android/mobly/snippet/bundled/utils/Utils.java#L205)):*
> ```Java
>      * @return a String with the ASCII hex representation.
>      */
>     public static String bytesToHex(byte[] bytes) {
> ```

`bytesToHexString`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80)*

kdart:
Review status: 0 of 3 files reviewed at latest revision, 5 unresolved discussions.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 84 at r3](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-Kv-asBcqwFKT3J8DZjo:-Kv0WWpuYzOYVml321rt:b-896fix) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/0c43c4fa602b08a96bfc54a541cd48c7df111aeb/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L84)):*

we should throw exception instead of returning null
</blockquote></details>

Done.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/utils/Utils.java, line 205 at r3](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-Kv-bGEZxGip9vyOMZ8O:-Kv0WYMlrYm2bUxWnRcK:b-896fix) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/0c43c4fa602b08a96bfc54a541cd48c7df111aeb/src/main/java/com/google/android/mobly/snippet/bundled/utils/Utils.java#L205)):*

`bytesToHexString`
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80)*

xpconanfan:
Review status: 0 of 3 files reviewed at latest revision, 9 unresolved discussions.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 132 at r2](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-KuzPGz_30TLm2csrY0P:-Kv19wriwPkzntLpQ2Bc:b69ufhn) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/ed10e12060de46fdec39643f90311779e80d22b8/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L132)):*

i wouldn't bet on it.
any app could download anything at any time during the test on a device.
you don't want this to be mistakenly flipped, which will make the test flaky.

Let's figure out a way.
</blockquote></details>

Maybe the request ID is part of this intent?

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 90 at r4](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-Kv18ckF4AFzQzsKQxfe:-Kv18ckF4AFzQzsKQxff:b5zjh5z) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/a56479f0f9e12d681d0301f4c8f1dc82fc8b751c/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L90)):*
> ```Java
>         if (pathsegments.size() < 1) {
>             String msg = String.format("The Uri %s does not have a path.", uri.toString());
>             Log.d(msg);
> ```

It's redundant to log the same message as it will be logged as part of the throwing.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 120 at r4](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-Kv19HmGwgABKJ5T8-dd:-Kv19HmGwgABKJ5T8-de:b-pmmhv4) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/a56479f0f9e12d681d0301f4c8f1dc82fc8b751c/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L120)):*
> ```Java
>     }
> 
>     @Rpc(description = "Compute MD5 hash on a content URI.")
> ```

this is not specific to networking right?
We could do:
`FileSnippet#fileCalculateMd5Hash`

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 121 at r4](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-Kv19LxFovbvrcJR2M4_:-Kv19LxFovbvrcJR2M4a:b-7to9w1) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/a56479f0f9e12d681d0301f4c8f1dc82fc8b751c/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L121)):*
> ```Java
> 
>     @Rpc(description = "Compute MD5 hash on a content URI.")
>     public String networkMD5Hash(String url) throws IOException, NoSuchAlgorithmException  {
> ```

`uri`

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 122 at r4](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-Kv19TdzpvXu9nqo2I_i:-Kv19TdzpvXu9nqo2I_j:b-2jsqhv) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/a56479f0f9e12d681d0301f4c8f1dc82fc8b751c/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L122)):*
> ```Java
>     @Rpc(description = "Compute MD5 hash on a content URI.")
>     public String networkMD5Hash(String url) throws IOException, NoSuchAlgorithmException  {
>         ParcelFileDescriptor pfd;
> ```

why are these local vars declared in the front but others are not?
Let's inline them?

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 124 at r4](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-Kv19k5m1ElJaX7IY3Lx:-Kv19k5m1ElJaX7IY3Ly:b-412egk) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/a56479f0f9e12d681d0301f4c8f1dc82fc8b751c/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L124)):*
> ```Java
>         ParcelFileDescriptor pfd;
>         MessageDigest md;
>         String hexdigest = "";
> ```

`md5String`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80)*

kdart:
Review status: 0 of 3 files reviewed at latest revision, 9 unresolved discussions.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 132 at r2](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-KuzPGz_30TLm2csrY0P:-Kv4H-GRzZsssXd5jUTN:bikcjlr) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/ed10e12060de46fdec39643f90311779e80d22b8/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L132)):*

Maybe the request ID is part of this intent?
</blockquote></details>

After some digging it turns out it does. 8-)

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 90 at r4](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-Kv18ckF4AFzQzsKQxfe:-Kv4DhS29ONMq7lF-zeC:b-896fix) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/a56479f0f9e12d681d0301f4c8f1dc82fc8b751c/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L90)):*

It's redundant to log the same message as it will be logged as part of the throwing.
</blockquote></details>

Done.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 120 at r4](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-Kv19HmGwgABKJ5T8-dd:-Kv4HDcA8tr4m5U_W6tu:boxan7e) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/a56479f0f9e12d681d0301f4c8f1dc82fc8b751c/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L120)):*

this is not specific to networking right?
We could do:
`FileSnippet#fileCalculateMd5Hash`
</blockquote></details>

In this case the parameter is a content URI associated with downloads. We can generalize and make a new File module later. For now, this unblocks what I'm trying to unblock. 

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 121 at r4](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-Kv19LxFovbvrcJR2M4_:-Kv4EFexYoufrgDZbh9r:bs2s85p) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/a56479f0f9e12d681d0301f4c8f1dc82fc8b751c/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L121)):*

`uri`
</blockquote></details>

Done. There is another local var with same name. Also changed that var name.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 122 at r4](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-Kv19TdzpvXu9nqo2I_i:-Kv4EcKdvvIKZD7zuVOP:b-896fix) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/a56479f0f9e12d681d0301f4c8f1dc82fc8b751c/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L122)):*

why are these local vars declared in the front but others are not?
Let's inline them?
</blockquote></details>

Done.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 124 at r4](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-Kv19k5m1ElJaX7IY3Lx:-Kv4EmsbghTdHg_Ucc-O:b-pnipy6) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/a56479f0f9e12d681d0301f4c8f1dc82fc8b751c/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L124)):*

`md5String`
</blockquote></details>

Done. However, it is a message digest that is a hex string.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80)*

kdart:
new Usage:
> uri = s.networkHTTPDownload("http://www.google.com/images/branding/googleg/1x/googleg_standard_color_128dp.png")
> uri
> 'content://downloads/all_downloads/102'
> s.networkMD5Hash(uri)
> '116d7abaea20868404cb366b611c69b7'

xpconanfan:
Review status: 0 of 3 files reviewed at latest revision, 7 unresolved discussions.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 120 at r4](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-Kv19HmGwgABKJ5T8-dd:-Kv4eoii86cVkHtouDdd:b-p5gz1u) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/a56479f0f9e12d681d0301f4c8f1dc82fc8b751c/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L120)):*
<details><summary><i>Previously, kdart (Keith Dart) wrote…</i></summary><blockquote>

In this case the parameter is a content URI associated with downloads. We can generalize and make a new File module later. For now, this unblocks what I'm trying to unblock. 
</blockquote></details>

`uri` is a standard way of referring to files in Android, it's not specific to network download.
So it's strange to have this in network snippet.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 85 at r5](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-Kv4ShzGrmlvDiLenRt-:-Kv4ShzGrmlvDiLenRt0:bx67wgt) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/69a7506d272df7b3d64537e35ef9547aad3c7971/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L85)):*
> ```Java
>     @Rpc(description = "Download a file using HTTP. Return content Uri (file remains on device). "
>                        + "The Uri should be treated as an opaque handle for further operations.")
>     public String networkHTTPDownload(String url) throws IllegalArgumentException, NetworkingSnippetException {
> ```

`networkHttpDownload`

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 121 at r5](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-Kv4elBmLSy7YP6Gf8pS:-Kv4elBmLSy7YP6Gf8pT:b4mkap3) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/69a7506d272df7b3d64537e35ef9547aad3c7971/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L121)):*
> ```Java
> 
>     @Rpc(description = "Compute MD5 hash on a content URI. Return the MD5 has has a hex string.")
>     public String networkMD5Hash(String uri) throws IOException, NoSuchAlgorithmException  {
> ```

`networkMd5Hash`

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 133 at r5](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-Kv4e9AoMihon2Lt4ZWQ:-Kv4e9AoMihon2Lt4ZWR:b-ww5yu4) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/69a7506d272df7b3d64537e35ef9547aad3c7971/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L133)):*
> ```Java
>         try {
>             dis.read(buf, 0, length);
>             md5string = Utils.bytesToHexString(md.digest());
> ```

seems like you can simply return here and do away with the local var?

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 148 at r5](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-Kv4eHn6X_Qj2WMgelMU:-Kv4eHn6X_Qj2WMgelMV:bb97m5s) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/69a7506d272df7b3d64537e35ef9547aad3c7971/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L148)):*
> ```Java
>             long gotid = (long) intent.getExtras().get("extra_download_id");
>             if (DownloadManager.ACTION_DOWNLOAD_COMPLETE.equals(action)
>                             && gotid == mReqid) {
> ```

We should probably pass the req ID to the constructor of `DownloadReceiver` so it's better isolated.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 149 at r5](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-Kv4e_W5bfXbeMMr6XQu:-Kv4e_W5bfXbeMMr6XQv:bomctar) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/69a7506d272df7b3d64537e35ef9547aad3c7971/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L149)):*
> ```Java
>             if (DownloadManager.ACTION_DOWNLOAD_COMPLETE.equals(action)
>                             && gotid == mReqid) {
>                 mIsDownloadComplete = true;
> ```

do we want to handle multiple calls to the download Rpc?
If so, we should probably make a `Map<ReqId, Boolean>` to track the status of each call?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80)*

kdart:
Review status: 0 of 3 files reviewed at latest revision, 7 unresolved discussions.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 120 at r4](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-Kv19HmGwgABKJ5T8-dd:-Kv4nTJnAv5bmpDHw9ar:b-gkzap0) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/a56479f0f9e12d681d0301f4c8f1dc82fc8b751c/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L120)):*

`uri` is a standard way of referring to files in Android, it's not specific to network download.
So it's strange to have this in network snippet.
</blockquote></details>

Ok, ok, FilesSnippet it is. 

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 85 at r5](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-Kv4ShzGrmlvDiLenRt-:-Kv4nPCuWOY8eHERlbqP:b-896fix) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/69a7506d272df7b3d64537e35ef9547aad3c7971/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L85)):*

`networkHttpDownload`
</blockquote></details>

Done.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 121 at r5](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-Kv4elBmLSy7YP6Gf8pS:-Kv4nYRVX5m-PDDRIapw:b-896fix) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/69a7506d272df7b3d64537e35ef9547aad3c7971/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L121)):*

`networkMd5Hash`
</blockquote></details>

Done.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 133 at r5](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-Kv4e9AoMihon2Lt4ZWQ:-Kv4n_CEW-AD6WZneZvP:b-896fix) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/69a7506d272df7b3d64537e35ef9547aad3c7971/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L133)):*

seems like you can simply return here and do away with the local var?
</blockquote></details>

Done.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 148 at r5](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-Kv4eHn6X_Qj2WMgelMU:-Kv4oUiN2Vj1n8xShpHQ:bo0ngdd) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/69a7506d272df7b3d64537e35ef9547aad3c7971/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L148)):*

We should probably pass the req ID to the constructor of `DownloadReceiver` so it's better isolated.
</blockquote></details>

The DownloadReceiver is constructed and registered before the request id is known. I think trying to do it after would introduce a race.  I think this is fine for now. It does what we currently need it to do. 

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 149 at r5](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-Kv4e_W5bfXbeMMr6XQu:-Kv4nm57pLAj1h49avd9:b-83gdpv) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/69a7506d272df7b3d64537e35ef9547aad3c7971/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L149)):*

do we want to handle multiple calls to the download Rpc?
If so, we should probably make a `Map<ReqId, Boolean>` to track the status of each call?
</blockquote></details>

This could really be a future feature, when the need arises. This method blocks (due to WaitUntil) so it won't be an issue for the current client using this, and there are always just one of those. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80)*

kdart:
New usage:

> uri = s.networkHttpDownload("http://www.google.com/images/branding/googleg/1x/googleg_standard_color_128dp.png")
> uri
> 'content://downloads/all_downloads/1'
> s.fileMd5Hash(uri)
> '116d7abaea20868404cb366b611c69b7'

xpconanfan:
Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 99 at r6](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-Kv8yTroi_zd7Y1vA-lM:-Kv8yTroi_zd7Y1vA-lN:b4c4w9e) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/1dc32d92bf0d2337e3a2bfe8fb41ea0244b0b54a/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L99)):*
> ```Java
>         try {
>             mReqid = mDownloadManager.enqueue(request);
>             Log.d(String.format("networkHTTPDownload download of %s with id %d", url, mReqid));
> ```

Why do we need this token "networkHTTPDownload"?

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 131 at r6](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-Kv8zCTYbkM7rmYelugz:-Kv8zCTZTAhwuNeBdmTe:b-q6loem) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/1dc32d92bf0d2337e3a2bfe8fb41ea0244b0b54a/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L131)):*
> ```Java
> 
>     @Override
>     public void shutdown() {}
> ```

should shutdown kill any remaining downloads?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80)*

kdart:
Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 99 at r6](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-Kv8yTroi_zd7Y1vA-lM:-Kv9JaLiIPRBcylVOGBu:b-63l0zb) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/1dc32d92bf0d2337e3a2bfe8fb41ea0244b0b54a/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L99)):*

Why do we need this token "networkHTTPDownload"?
</blockquote></details>

I like putting the method names in log messages to make it clearer where the log came from. 

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 131 at r6](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-Kv8zCTYbkM7rmYelugz:-Kv9JjmWtxCar7swd70z:besi4m6) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/1dc32d92bf0d2337e3a2bfe8fb41ea0244b0b54a/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L131)):*

should shutdown kill any remaining downloads?
</blockquote></details>

The DownloadManager is asynchronous so I think we can just let it finish. It won't hurt anything. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80)*

xpconanfan:
Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 99 at r6](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-Kv8yTroi_zd7Y1vA-lM:-Kv9MHfp1JT590tjn5hD:b-r3nnde) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/1dc32d92bf0d2337e3a2bfe8fb41ea0244b0b54a/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L99)):*
<details><summary><i>Previously, kdart (Keith Dart) wrote…</i></summary><blockquote>

I like putting the method names in log messages to make it clearer where the log came from. 
</blockquote></details>

Ok, let's make it match the method name then

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 131 at r6](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-Kv8zCTYbkM7rmYelugz:-Kv9MXO4hA7-bYjk45G3:b-2uu45h) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/1dc32d92bf0d2337e3a2bfe8fb41ea0244b0b54a/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L131)):*
<details><summary><i>Previously, kdart (Keith Dart) wrote…</i></summary><blockquote>

The DownloadManager is asynchronous so I think we can just let it finish. It won't hurt anything. 
</blockquote></details>

It's not good to leave zombie processes going on.
It's a time bomb for unknown behavior and racing conditions.
Let's clean it up.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80)*

kdart:
Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 99 at r6](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-Kv8yTroi_zd7Y1vA-lM:-Kv9Rb2yXMMUzdcQuefl:b-896fix) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/1dc32d92bf0d2337e3a2bfe8fb41ea0244b0b54a/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L99)):*

Ok, let's make it match the method name then
</blockquote></details>

Done.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 131 at r6](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-Kv8zCTYbkM7rmYelugz:-Kv9RdnKP9Wp1R6C5BZW:b-hwoj2l) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/1dc32d92bf0d2337e3a2bfe8fb41ea0244b0b54a/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L131)):*

It's not good to leave zombie processes going on.
It's a time bomb for unknown behavior and racing conditions.
Let's clean it up.
</blockquote></details>

It's not a zombie process, though. It's an en-queued  job for the persistent download manager. Not our process. 
It simply sends the DOWNLOAD_COMPLETE, and we aren't there to get the broadcast then it's no big deal. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80)*

xpconanfan:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 131 at r6](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-Kv8zCTYbkM7rmYelugz:-Kv9Sf6JCg6t9zf-PvTe:b2vpj6y) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/1dc32d92bf0d2337e3a2bfe8fb41ea0244b0b54a/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L131)):*
<details><summary><i>Previously, kdart (Keith Dart) wrote…</i></summary><blockquote>

It's not a zombie process, though. It's an en-queued  job for the persistent download manager. Not our process. 
It simply sends the DOWNLOAD_COMPLETE, and we aren't there to get the broadcast then it's no big deal. 

</blockquote></details>

Well, it's running somewhere ;)
Can we simply call `DownloadManager.remove` with the req id to tell it to clean up?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80)*

xpconanfan:
Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 99 at r6](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-Kv8yTroi_zd7Y1vA-lM:-Kv9SvYOk4yKZXGEyplU:btqa3k0) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/1dc32d92bf0d2337e3a2bfe8fb41ea0244b0b54a/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L99)):*
<details><summary><i>Previously, kdart (Keith Dart) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

Hmm, I still see "HTTP" in R6, it should be "Http"?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80)*

kdart:
Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 99 at r6](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-Kv8yTroi_zd7Y1vA-lM:-Kv9U8dCORbGJL48mZvK:b-mhcamc) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/1dc32d92bf0d2337e3a2bfe8fb41ea0244b0b54a/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L99)):*

Hmm, I still see "HTTP" in R6, it should be "Http"?
</blockquote></details>

one sec...

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 131 at r6](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-Kv8zCTYbkM7rmYelugz:-Kv9UDMJzfMbfHey1dyT:b-40mu3s) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/1dc32d92bf0d2337e3a2bfe8fb41ea0244b0b54a/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L131)):*

Well, it's running somewhere ;)
Can we simply call `DownloadManager.remove` with the req id to tell it to clean up?
</blockquote></details>

I suppose we could do that, however it's also most likely that it had already completed. 
If it becomes a problem, which I doubt, we can add it then.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80)*

xpconanfan:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 131 at r6](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-Kv8zCTYbkM7rmYelugz:-Kv9WPxbJswej4rM5RK6:bh077dm) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/1dc32d92bf0d2337e3a2bfe8fb41ea0244b0b54a/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L131)):*
<details><summary><i>Previously, kdart (Keith Dart) wrote…</i></summary><blockquote>

I suppose we could do that, however it's also most likely that it had already completed. 
If it becomes a problem, which I doubt, we can add it then.

</blockquote></details>

Can we add the call here then?
So there's no stone left unturned ;)

When multi-threading is involved, "almost certain" has burned me too many times (you know the snippet lib socket thing...)

Looks like `remove` will even remove the downloaded cache file, which is also very useful.
In my lab, we got a lot of device out of storage issues. So making sure cache file is removed is essential.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80)*

xpconanfan:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 131 at r6](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-Kv8zCTYbkM7rmYelugz:-Kv9XCSTgfDwYRsQzpvT:b-wa4e6z) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/1dc32d92bf0d2337e3a2bfe8fb41ea0244b0b54a/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L131)):*

Can we add the call here then?
So there's no stone left unturned ;)

When multi-threading is involved, "almost certain" has burned me too many times (you know the snippet lib socket thing...)

Looks like `remove` will even remove the downloaded cache file, which is also very useful.
In my lab, we got a lot of device out of storage issues. So making sure cache file is removed is essential.
</blockquote></details>

we could add an explicit Rpc to call `remove` and call it in test tear down.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80)*

kdart:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 131 at r6](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-Kv8zCTYbkM7rmYelugz:-Kv9XQiLsYyt_vxAxDru:b-i8p6x2) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/1dc32d92bf0d2337e3a2bfe8fb41ea0244b0b54a/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L131)):*

we could add an explicit Rpc to call `remove` and call it in test tear down.
</blockquote></details>

Ah, that is a good idea.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80)*

kdart:
New usage:
> uri = s.networkHttpDownload("http://www.google.com/images/branding/googleg/1x/googleg_standard_color_128dp.png")
> uri
> 'content://downloads/all_downloads/59'
> s.fileMd5Hash(uri)
> '116d7abaea20868404cb366b611c69b7'
> s.fileDeleteContent(uri)
1
> s.fileMd5Hash(uri)
> Traceback (most recent call last):  ...

kdart:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java, line 131 at r6](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-Kv8zCTYbkM7rmYelugz:-Kv9sCiEHxY_c9mxqltm:b-z0g2q6) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/1dc32d92bf0d2337e3a2bfe8fb41ea0244b0b54a/src/main/java/com/google/android/mobly/snippet/bundled/NetworkingSnippet.java#L131)):*
<details><summary><i>Previously, kdart (Keith Dart) wrote…</i></summary><blockquote>

Ah, that is a good idea.
</blockquote></details>

Done. Placed in FileSnippet module.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80)*

xpconanfan:
Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/FileSnippet.java, line 69 at r7](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-Kv9sRyvNjF_45hcrw6Q:-Kv9sRyweea2X8Y-xIUm:b-hqphun) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/06171a671fc89e496cf45849f7acbe181c33e39f/src/main/java/com/google/android/mobly/snippet/bundled/FileSnippet.java#L69)):*
> ```Java
>     @Rpc(description = "Remove a file pointed to by the content URI. Return "
>             + "number deleted (should always be one)")
>     public int fileDeleteContent(String uri) {
> ```

Isn't this removing the file directly?
So maybe `fileDelete` is a better name?

`fileDeleteContent` sounds like it's clearing the content but keeping the file.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/FileSnippet.java, line 71 at r7](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-Kv9scTa3wc7JQcrg5ZO:-Kv9scTa3wc7JQcrg5ZP:b36beay) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/06171a671fc89e496cf45849f7acbe181c33e39f/src/main/java/com/google/android/mobly/snippet/bundled/FileSnippet.java#L71)):*
> ```Java
>     public int fileDeleteContent(String uri) {
>         Uri uri_ = Uri.parse(uri);
>         return mContext.getContentResolver().delete(uri_, null, null);
> ```

why return this if it's always 1?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80)*

kdart:
Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/FileSnippet.java, line 69 at r7](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-Kv9sRyvNjF_45hcrw6Q:-Kv9swXX065ZpOdJqjRo:bjoihxi) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/06171a671fc89e496cf45849f7acbe181c33e39f/src/main/java/com/google/android/mobly/snippet/bundled/FileSnippet.java#L69)):*

Isn't this removing the file directly?
So maybe `fileDelete` is a better name?

`fileDeleteContent` sounds like it's clearing the content but keeping the file.
</blockquote></details>

The parameter is a content URI, not a file name. It goes through the abstract content manager. The name makes it more clear that we are deleting "content", which may or may not be an actual file (it could be a BLOB). 

---

*[src/main/java/com/google/android/mobly/snippet/bundled/FileSnippet.java, line 71 at r7](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-Kv9scTa3wc7JQcrg5ZO:-Kv9tJx3y6XBa-fGvxSo:b-awh4ip) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/06171a671fc89e496cf45849f7acbe181c33e39f/src/main/java/com/google/android/mobly/snippet/bundled/FileSnippet.java#L71)):*

why return this if it's always 1?
</blockquote></details>

Because we can. :-) It can be used to verify that we got one entry, or you can ignore it. Other kinds of URIs might return more than one, but I don't know. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80)*

xpconanfan:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/FileSnippet.java, line 71 at r7](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-Kv9scTa3wc7JQcrg5ZO:-Kv9vGHPs1rixiI2K2g3:bj59kt2) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/06171a671fc89e496cf45849f7acbe181c33e39f/src/main/java/com/google/android/mobly/snippet/bundled/FileSnippet.java#L71)):*
<details><summary><i>Previously, kdart (Keith Dart) wrote…</i></summary><blockquote>

Because we can. :-) It can be used to verify that we got one entry, or you can ignore it. Other kinds of URIs might return more than one, but I don't know. 
</blockquote></details>

If we only expect the value to be 1, maybe it's more useful to assert that it is 1 and throw an error otherwise.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80)*

kdart:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/FileSnippet.java, line 71 at r7](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-Kv9scTa3wc7JQcrg5ZO:-Kv9vmH-U2Ukm5jVDtFD:b-3aktpb) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/06171a671fc89e496cf45849f7acbe181c33e39f/src/main/java/com/google/android/mobly/snippet/bundled/FileSnippet.java#L71)):*

If we only expect the value to be 1, maybe it's more useful to assert that it is 1 and throw an error otherwise.
</blockquote></details>

I intended to do that on the Python side. Keep the server side simpler and more general. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80)*

kdart:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/FileSnippet.java, line 71 at r7](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-Kv9scTa3wc7JQcrg5ZO:-Kv9wI2z6Tx0LzFXfxbK:b-bxx1kw) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/06171a671fc89e496cf45849f7acbe181c33e39f/src/main/java/com/google/android/mobly/snippet/bundled/FileSnippet.java#L71)):*
<details><summary><i>Previously, kdart (Keith Dart) wrote…</i></summary><blockquote>

I intended to do that on the Python side. Keep the server side simpler and more general. 
</blockquote></details>

BTW, it keeps the RPC API more consistent with the underlaying content API. I think it's generally better to keep this RPC as thin as possible. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80)*

xpconanfan:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/FileSnippet.java, line 71 at r7](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-Kv9scTa3wc7JQcrg5ZO:-Kv9wGkUjopYA_ZNiI6Z:b-f58yr6) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/06171a671fc89e496cf45849f7acbe181c33e39f/src/main/java/com/google/android/mobly/snippet/bundled/FileSnippet.java#L71)):*
<details><summary><i>Previously, kdart (Keith Dart) wrote…</i></summary><blockquote>

BTW, it keeps the RPC API more consistent with the underlaying content API. I think it's generally better to keep this RPC as thin as possible. 
</blockquote></details>

In the new world we actually prefer to keep logic on the Java side.
This is especially the case in MBS, e.g. wifiConnect checks a lot of conditions and throw errors.

The Rpc says it deletes a content, so it should either delete a content or throw an error if something went wrong.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80)*

kdart:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/FileSnippet.java, line 71 at r7](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-Kv9scTa3wc7JQcrg5ZO:-Kv9xIdNNE6g10UmL3xC:b-gof6ie) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/06171a671fc89e496cf45849f7acbe181c33e39f/src/main/java/com/google/android/mobly/snippet/bundled/FileSnippet.java#L71)):*

In the new world we actually prefer to keep logic on the Java side.
This is especially the case in MBS, e.g. wifiConnect checks a lot of conditions and throw errors.

The Rpc says it deletes a content, so it should either delete a content or throw an error if something went wrong.
</blockquote></details>

They underlaying API already throws an error if something went wrong, which is passed through to the client. 

I also verified that it does delete the file. The file does end up in /sdard/Downloads on my device. The file is removed after that call. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80)*

xpconanfan:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/FileSnippet.java, line 71 at r7](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-Kv9scTa3wc7JQcrg5ZO:-Kv9ys_3BiTKkZjO8H8G:b-4y5kv8) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/06171a671fc89e496cf45849f7acbe181c33e39f/src/main/java/com/google/android/mobly/snippet/bundled/FileSnippet.java#L71)):*
<details><summary><i>Previously, kdart (Keith Dart) wrote…</i></summary><blockquote>

They underlaying API already throws an error if something went wrong, which is passed through to the client. 

I also verified that it does delete the file. The file does end up in /sdard/Downloads on my device. The file is removed after that call. 

</blockquote></details>

Then what do you want to verify on the Python side with the ret value?

If nothing, let's remove the return value.
Otherwise let's assert the ret value is 1.

How is a user supposed to figure out what a random ret value of 1 means?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80)*

xpconanfan:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/FileSnippet.java, line 71 at r7](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-Kv9scTa3wc7JQcrg5ZO:-Kv9zEB6sjaCHZW7N_-0:b9bqksw) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/06171a671fc89e496cf45849f7acbe181c33e39f/src/main/java/com/google/android/mobly/snippet/bundled/FileSnippet.java#L71)):*

Then what do you want to verify on the Python side with the ret value?

If nothing, let's remove the return value.
Otherwise let's assert the ret value is 1.

How is a user supposed to figure out what a random ret value of 1 means?
</blockquote></details>

* Otherwise let's assert the value is 1 in java

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80)*

kdart:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/main/java/com/google/android/mobly/snippet/bundled/FileSnippet.java, line 71 at r7](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-Kv9scTa3wc7JQcrg5ZO:-KvA09Q-eY2oBcxWnves:b-ykj2b2) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/06171a671fc89e496cf45849f7acbe181c33e39f/src/main/java/com/google/android/mobly/snippet/bundled/FileSnippet.java#L71)):*

* Otherwise let's assert the value is 1 in java
</blockquote></details>

Done. Just removed it. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80)*

xpconanfan:
:lgtm:

I assume the linter was run as well?

---

Review status: 0 of 4 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-:-KvA0utNzRMyW5R3nG_F:b-s1r1zw)*

kdart:
Yes, nothing pertinent to this change. lint does report some pre-existing errors. 

---

Review status: 0 of 4 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/google/mobly-bundled-snippets/80#-:-KvA3fPnDCe3d1NEvAGj:bhimgpx)*

PR #114, opened 2019-06-24
Cleanup UiAutomator connection

k2fong:
cleanup UiAutomator connection after use.  This will enable other UiAutomator user to properly connect instead of getting an "already registered" exception because UiAutomator only accept a single connection at any giving time.

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly-bundled-snippets/114)


k2fong:
Verified locally that wifi disable/enable still functions properly after disconnecting from uiautomator on Q

k2fong:
sorry about the latest update after review, made a small change to more accurately check when the uiautomation permission needs to be adopted.

https://developer.android.com/preview/privacy/camera-connectivity#target-q

PR #125, opened 2020-01-08
Fix exception in "trimQuotationMarks()" method

ryancllin:
Add a string length check on the given param to avoid
StringIndexOutOfBoundsException.

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly-bundled-snippets/125)

### DIFF
--- a/src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java
+++ b/src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java
@@ -62,7 +62,8 @@ public class JsonSerializer {
      */
     public static String trimQuotationMarks(String originalString) {
         String result = originalString;
-        if (originalString.charAt(0) == '"'
+        if (originalString.length() > 2
+                && originalString.charAt(0) == '"'
                 && originalString.charAt(originalString.length() - 1) == '"') {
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java
```diff
@@ -62,7 +62,8 @@ public JsonSerializer() {
      */
     public static String trimQuotationMarks(String originalString) {
         String result = originalString;
-        if (originalString.charAt(0) == '"'
+        if (originalString.length() > 2
```
xpconanfan: what should we do if the string length is <= 2?

ryancllin: If the string length is <=2, it will return it directly (as it is).
A minimum and meaningful input string like "a" (len = 3) will be trimmed out double quotation marks successfuly. For other inputs(with or without quotation marks), this method will return them without any change.
Currently, this method is mainly used to deal with quotation marks of a wifi SSID string. And this bug fix try to solve an exception case. I think let its upper caller to deal with a null/empty wifi SSID string case is okay. It doesn't need to return something like "invalid SSID string" here.
WDYT? :)

PR #146, opened 2022-01-11
Have wifiClearConfiguredNetworks re-check network list to determine operation success.

xianyuanjia:
Sometimes, a removeNetwork call could remove more than one configured network at a time (e.g. a network and its OWE config), so that the subsequent call may find its target config missing and return false. Currently, this will cause an exception to be raised, even though the intended outcome is achieved. Therefore, it is more reliable to simply verify that the configured network list is cleared by the end of the removal process.

---
This change is [Reviewable](https://reviewable.io/reviews/google/mobly-bundled-snippets/146)

### DIFF
--- a/src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
+++ b/src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
@@ -78,15 +78,25 @@ public class WifiManagerSnippet implements Snippet {
                             + "networks were added through this MBS instance")
     public void wifiClearConfiguredNetworks() throws WifiManagerSnippetException {
         List<WifiConfiguration> unremovedConfigs = mWifiManager.getConfiguredNetworks();
+        List<WifiConfiguration> failedConfigs = new ArrayList<>();
         if (unremovedConfigs == null) {
             throw new WifiManagerSnippetException(
                     "Failed to get a list of configured networks. Is wifi disabled?");
         }
         for (WifiConfiguration config : unremovedConfigs) {
             if (!mWifiManager.removeNetwork(config.networkId)) {
-                Log.e("Encountered error while removing network: " + config);
+                failedConfigs.add(config);
             }
         }
+
+        // If removeNetwork is called on a network with both an open and OWE config, it will remove
+        // both. The subsequent call on the same network will fail. The clear operation may succeed
+        // even if failures appear in the log below.
+        if (!failedConfigs.isEmpty()) {
+            Log.e("Encountered error while removing networks: " + failedConfigs);
+        }
+
+        // Re-check configured configs list to ensure that it is cleared
         unremovedConfigs = mWifiManager.getConfiguredNetworks();
         if (!unremovedConfigs.isEmpty()) {
             throw new WifiManagerSnippetException("Failed to remove networks: " + unremovedConfigs);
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
```diff
@@ -78,18 +78,18 @@ public WifiManagerSnippet() throws Throwable {
...
         List<WifiConfiguration> unremovedConfigs = mWifiManager.getConfiguredNetworks();
-        List<WifiConfiguration> failedConfigs = new ArrayList<>();
         if (unremovedConfigs == null) {
             throw new WifiManagerSnippetException(
                     "Failed to get a list of configured networks. Is wifi disabled?");
         }
         for (WifiConfiguration config : unremovedConfigs) {
             if (!mWifiManager.removeNetwork(config.networkId)) {
-                failedConfigs.add(config);
+                Log.e("Encountered error while removing network: " + config);
```
xpconanfan, line 87: let's still keep the failedConfigs list and log the list altogether if the clear failed at the end?

xpconanfan (changes requested):

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
```diff
@@ -78,18 +78,18 @@ public WifiManagerSnippet() throws Throwable {
                             + "networks were added through this MBS instance")
     public void wifiClearConfiguredNetworks() throws WifiManagerSnippetException {
         List<WifiConfiguration> unremovedConfigs = mWifiManager.getConfiguredNetworks();
-        List<WifiConfiguration> failedConfigs = new ArrayList<>();
         if (unremovedConfigs == null) {
             throw new WifiManagerSnippetException(
                     "Failed to get a list of configured networks. Is wifi disabled?");
         }
         for (WifiConfiguration config : unremovedConfigs) {
             if (!mWifiManager.removeNetwork(config.networkId)) {
```
xpconanfan, line 86: we should probably add some context with inline comments on the new strategy.
this is because Wi-Fi stack has special types of networks (what's the official term?) that come and go on their own right?

xpconanfan (changes requested):

xianyuanjia (commented):
> *[Reviewable](https://reviewable.io/reviews/google/mobly-bundled-snippets/146)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @xianyuanjia and @xpconanfan)

___
*[src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java, line 86 at r1](https://reviewable.io/reviews/google/mobly-bundled-snippets/146#-MtAJOTGqVRUawIAZb1a-r1-86:-MtLMctEBSEtauP8RkFU:b-896fix) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/abb239def342d958aec3d2fad501aca599d83cc0/src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java#L86)):*

we should probably add some context with inline comments on the new strategy.
this is because Wi-Fi stack has special types of networks (what's the official term?) that come and go on their own right?
</blockquote></details>

Done.
___
*[src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java, line 87 at r1](https://reviewable.io/reviews/google/mobly-bundled-snippets/146#-MtAJOTGqVRUawIAZb1a-r1-87:-MtLMdknDR4-6XHWSkuZ:b-896fix) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/abb239def342d958aec3d2fad501aca599d83cc0/src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java#L87)):*

let's still keep the failedConfigs list and log the list altogether if the clear failed at the end?
</blockquote></details>

Done.

xpconanfan (approved):

xianyuanjia (commented):
> Reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/mobly-bundled-snippets/146)* status: all files reviewed, 2 unresolved discussions (waiting on @xpconanfan)

PR #149, opened 2022-05-04
Use intent to enable BT discoverability for T+; handle dialog with UIAutomator

xianyuanjia:
The "BluetoothAdapter#setScanMode" API is no longer accessible to SDK versions 31 and above. Add a intent-based workflow according to https://developer.android.com/reference/android/bluetooth/BluetoothAdapter#ACTION_REQUEST_DISCOVERABLE. Use UI Automator to interact with the pop-up user dialog.

This change is [Reviewable](https://reviewable.io/reviews/google/mobly-bundled-snippets/149)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -34,7 +34,7 @@ android {
 
     defaultConfig {
         applicationId "com.google.android.mobly.snippet.bundled"
-        minSdkVersion 15
+        minSdkVersion 18
         // Set target to 22 to avoid having to deal with runtime permissions.
         targetSdkVersion 22
         versionCode 1
@@ -72,6 +72,7 @@ artifacts {
 
 dependencies {
     implementation 'androidx.test:runner:1.3.0'
+    implementation 'com.android.support.test.uiautomator:uiautomator-v18:2.1.3'
     implementation 'com.google.android.mobly:mobly-snippet-lib:1.2.0'
     implementation 'com.google.code.gson:gson:2.8.6'
     implementation 'com.google.guava:guava:31.0.1-jre'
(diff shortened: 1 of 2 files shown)


xpconanfan (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/mobly-bundled-snippets/149)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @xianyuanjia and @xpconanfan)

___
*[`src/main/java/com/google/android/mobly/snippet/bundled/bluetooth/BluetoothAdapterSnippet.java` line 223 at r1](https://reviewable.io/reviews/google/mobly-bundled-snippets/149#-N1pjMOk1STza7ek-S-j:-N1pjMOk1STza7ek-S-k:b-2injib) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/2897fa78aba9d5be58873cd560bd26c341d3968f/src/main/java/com/google/android/mobly/snippet/bundled/bluetooth/BluetoothAdapterSnippet.java#L223)):*
> ```Java
>         }
>         if (Build.VERSION.SDK_INT > 30) {
>             // setScanMode is removed from public SDK for T and above, so uses an intent instead.
> ```

if this is removed, does that mean we can trigger a scan without calling this at all?
or has the discoverable flow changed completely (to intent)

xianyuanjia:
According to https://developer.android.com/reference/android/bluetooth/BluetoothAdapter#ACTION_REQUEST_DISCOVERABLE, the device is not discoverable by default. An intent is needed to allow scanning.

xpconanfan:
I'm not sure if `androidx.test.uiautomator` would cause build errors for internal users' snippets that use this lib.
So make sure we run global presubmit when syncing to internal.

xpconanfan (approved):

xianyuanjia:
The plan is to make string replacements during sync to set the supported package name.

PR #155, opened 2022-09-21
Add ScanRecord#getManufacturerSpecificData() to BLE Scan result.

ko1in1u:
This change is [Reviewable](https://reviewable.io/reviews/google/mobly-bundled-snippets/155)

### DIFF
--- a/src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java
+++ b/src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java
@@ -27,6 +27,7 @@ import android.net.wifi.WifiInfo;
 import android.os.Build;
 import android.os.Bundle;
 import android.os.ParcelUuid;
+import android.util.SparseArray;
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 import java.lang.reflect.Modifier;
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java
```diff
@@ -187,6 +188,20 @@ private Bundle serializeBleScanRecord(ScanRecord record) {
...
+    }
+
+    /** Serialize manufacturer specific data from ScanRecord for Bluetooth LE. */
+    @TargetApi(Build.VERSION_CODES.LOLLIPOP)
+    private Bundle serializeBleScanManufacturerSpecificData(ScanRecord record) {
+        Bundle result = new Bundle();
+        SparseArray<byte[]> sparseArray = record.getManufacturerSpecificData();
+        for (int i = 0; i < sparseArray.size(); i++) {
+            int key = sparseArray.keyAt(i);
+            result.putByteArray(String.valueOf(key), sparseArray.get(key));
```
xpconanfan, line 203: do we not need to encode the value which is of `byte` type?
because json doesn't support passing `byte`

ko1in1u: The values are all numbers and it can be passed by json in int type.

PR #156, opened 2023-01-19
Does not throw an error due to failure to remove Wi-Fi which were not added by MBS.

ko1in1u:
This change is [Reviewable](https://reviewable.io/reviews/google/mobly-bundled-snippets/156)

### DIFF
--- a/src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
+++ b/src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
@@ -22,6 +22,7 @@ import android.content.Context;
 import android.content.Intent;
 import android.content.IntentFilter;
 import android.net.wifi.ScanResult;
+import android.net.wifi.SupplicantState;
 import android.net.wifi.WifiConfiguration;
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
```diff
@@ -84,6 +82,10 @@ public void wifiClearConfiguredNetworks() throws WifiManagerSnippetException {
                     "Failed to get a list of configured networks. Is wifi disabled?");
         }
         for (WifiConfiguration config : unremovedConfigs) {
+            if (!config.toString().contains("cname=com.google.android.mobly.snippet.bundled")) {
```
xpconanfan, line 85: Identifying by package name is usually not great.
what if users build their own snippet apk and use this as a lib?

since all the networks are added via APIs in this class, is there a better way to identify them for removal?

xpconanfan (changes requested):

ko1in1u:
WifiConfiguration is deprecated in API level 29, discard this change.

PR #158, opened 2023-02-22
Modernise the build

yschimke:
I tried following the instructions, but after opening in Android Studio I got a bunch of errors.
But also jcenter is down.  So this was not building.

Happy to take guidance on what to fix.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/mobly-bundled-snippets/158)


yschimke:
I'm testing out these changes. Specifically I worry about some of the deprecated code and bumping the targetSdk.

        // Set target to 22 to avoid having to deal with runtime permissions.

xpconanfan:
Any update on the last comment?
What are the breakages from bumping the targetSDK? :)

yschimke:
I haven't observed any, but have only used 2 or 3 of the snippets.  

Is it worth us adding some instrumentation tests, potentially against FTL or emulators?  They could run without mobly, just calling the snippets.

xpconanfan:
@ko1in1u let's see if we can update all the build files except the SDK ver for now

ko1in1u:
MBS has beeb modernised in PR#166.

PR #175, opened 2023-08-25
Create BluetoothHeadsetSnippet

Karkeys360:
Updated bluetooth profile snippets to include BluetoothHeadsetSnippet.


xpconanfan:
This class does not follow the convention of MBS, so I'm not sure if it belongs here.
You are just using bt adapter APIs to perform a specific scenario.
MBS is meant to be a think layer that exposes Android APIs.

You can freely use any snippet class you have in conjunction with MBS without adding it to MBS.

xpconanfan:
Could you fix the google cla requirement?

PR #195, opened 2024-06-20
Add the network callback for wifi connection check

kaishi05:
Use the network callback for wifi connection status check for L+ devices.

Also remove networkId check for pre-L devices as it doesn't sometimes work while the network is already connected.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/mobly-bundled-snippets/195)

### DIFF
--- a/src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
+++ b/src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
@@ -59,47 +59,57 @@ public class WifiManagerSnippet implements Snippet {
 
     private static final int TIMEOUT_TOGGLE_STATE = 30;
     private final WifiManager mWifiManager;
+    private final ConnectivityManager mConnectivityManager;
     private final Context mContext;
     private final JsonSerializer mJsonSerializer = new JsonSerializer();
     private volatile boolean mIsScanResultAvailable = false;
-    private final AtomicBoolean isInternetWifiConnectedAtomic = new AtomicBoolean(false);
+    private final AtomicBoolean mIsWifiConnected = new AtomicBoolean(false);
 
     public WifiManagerSnippet() throws Throwable {
         mContext = InstrumentationRegistry.getInstrumentation().getContext();
         mWifiManager =
                 (WifiManager)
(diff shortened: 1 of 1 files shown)


xpconanfan (changes requested):
These RPCs are not specific to wifi network right?
Seem like they belong to the network snippet?

kaishi05:
Removed the internet check and only check if the wifi is available using networkCallback for L+ devices.. Removed networkId check in Pre-L devices.

Note that it is recommended by the framework to use networkCallback to check Wifi status. So this file is the right place to add the networkCallback to check Wifi status.

kaishi05:
Please  review 2nd comment.

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
```diff
@@ -373,9 +414,9 @@ public void wifiEnableSoftAp(@Nullable JSONObject configuration) throws Throwabl
             throw new WifiManagerSnippetException("Failed to initiate turning on Wi-Fi Soft AP.");
         }
         if (!Utils.waitUntil(() -> wifiIsApEnabled() == true, 60)) {
-            throw new WifiManagerSnippetException(
-                    "Timed out after 60s waiting for Wi-Fi Soft AP state to turn on with configuration: "
-                            + configuration);
+      throw new WifiManagerSnippetException(
+          "Timed out after 60s waiting for Wi-Fi Soft AP state to turn on with configuration: "
+              + configuration);
```
mhaoli, line 419: Could we revert the indent change in this PR? The modified lines look inconsistent with unmodified lines.

mhaoli (approved):
LGTM as this is the recommended WiFi status check approach from WiFi framework.

And this approach works for both WiFi network w/ and w/o internet access.

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
kaishi05: revert the unnecessary indent changes.

xpconanfan (approved):

PR #197, opened 2024-07-09
Update SSID check in WifiManagerSnippet.java

kaishi05:
getSSID() could return UNKNOWN_SSID when the location mode is disabled. Relax the SSID check to allow UNKNOWN_SSID so that test still passes when the location mode is disabled.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/mobly-bundled-snippets/197)

### DIFF
--- a/src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
+++ b/src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
@@ -109,7 +109,8 @@ public class WifiManagerSnippet implements Snippet {
     }
 
     private boolean isWifiConnectedToSsid(String ssid) {
-        return mWifiManager.getConnectionInfo().getSSID().equals(ssid);
+        return mWifiManager.getConnectionInfo().getSSID().equals(ssid)
+                || mWifiManager.getConnectionInfo().getSSID().equals(WifiManager.UNKNOWN_SSID);
     }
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
```diff
@@ -109,7 +109,8 @@ public boolean isWifiConnected() {
     }
 
     private boolean isWifiConnectedToSsid(String ssid) {
```
xpconanfan, line 111: The user specified an `ssid`, yet the second check will cause this to return true even if the connected ssid is not the one specified by user, which seems incorrect?

xpconanfan (changes requested):

src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
mhaoli: Looks like the new recommended get WifiInfo approach can get WifiInfo and ignore location sensitive data, `ConnectivityManager#getNetworkCapabilities(Network)#getTransportInfo()`.

So using this method would be better for this problem?

https://developer.android.com/reference/android/net/wifi/WifiManager#getConnectionInfo()

mhaoli (changes requested):

PR #199, opened 2024-08-01
Fix MBS Utils to work while being instrumented

Fozefy:
An instrumented context's getApplicationContext() may return null and throw.

In particular this can happen when using the BluetoothAdapterSnippet. Instead we get the applicationInfo from the packageManager.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/mobly-bundled-snippets/199)

### DIFF
--- a/src/main/java/com/google/android/mobly/snippet/bundled/utils/Utils.java
+++ b/src/main/java/com/google/android/mobly/snippet/bundled/utils/Utils.java
@@ -217,9 +217,7 @@ public final class Utils {
     }
 
    public static void adaptShellPermissionIfRequired(Context context) throws Throwable {
-      if (context.getPackageManager().getApplicationInfo(context.getPackageName(), 0).targetSdkVersion
-            >= 29
-          && Build.VERSION.SDK_INT >= 29) {
+      if (Build.VERSION.SDK_INT >= 29) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/android/mobly/snippet/bundled/utils/Utils.java
```diff
@@ -217,7 +217,8 @@ public static String bytesToHexString(byte[] bytes) {
     }
 
    public static void adaptShellPermissionIfRequired(Context context) throws Throwable {
-      if (context.getApplicationContext().getApplicationInfo().targetSdkVersion >= 29
+      if (context.getPackageManager().getApplicationInfo(context.getPackageName(), 0).targetSdkVersion
```
ko1in1u, line 220: I don't think we need to check the app's target SDK, not sure why it was added in the first place. Can you help remove it?

Fozefy: I've update the PR to simply remove the check. Do you generally squash commits in this repo, or can the changes go in like this?

xpconanfan: Thanks. We'll squash upon merging to master.

ko1in1u (approved):

PR #209, opened 2024-08-21
Rename RPC methods with wifi prefix.

kaishi05:
This change is [Reviewable](https://reviewable.io/reviews/google/mobly-bundled-snippets/209)

### DIFF
--- a/src/main/java/com/google/android/mobly/snippet/bundled/WifiAwareManagerSnippet.java
+++ b/src/main/java/com/google/android/mobly/snippet/bundled/WifiAwareManagerSnippet.java
@@ -34,13 +34,14 @@ public class WifiAwareManagerSnippet implements Snippet {
         }
     }
     private final Context mContext;
+    private boolean mIsAwareSupported;
     WifiAwareManager mWifiAwareManager;
 
     public WifiAwareManagerSnippet() throws Throwable {
         mContext = InstrumentationRegistry.getInstrumentation().getContext();
-        boolean isAwareSupported =
+        mIsAwareSupported =
             mContext.getPackageManager().hasSystemFeature(PackageManager.FEATURE_WIFI_AWARE);
-        if (isAwareSupported) {
+        if (mIsAwareSupported) {
             mWifiAwareManager = (WifiAwareManager) mContext.getSystemService(Context.WIFI_AWARE_SERVICE);
         }
         Utils.adaptShellPermissionIfRequired(mContext);
@@ -49,7 +50,7 @@ public class WifiAwareManagerSnippet implements Snippet {
     /** Checks if Aware is available. This could return false if WiFi or location is disabled. */
     @Rpc(description = "check if Aware is available.")
     public boolean wifiAwareIsAvailable() throws WifiAwareManagerSnippetException {
-        if (mWifiAwareManager == null) {
+        if (!mIsAwareSupported) {
             throw new WifiAwareManagerSnippetException(
                     "WifiAware is not supported in this device");
         }
--- a/src/main/java/com/google/android/mobly/snippet/bundled/WifiAwareManagerSnippet.java
+++ b/src/main/java/com/google/android/mobly/snippet/bundled/WifiAwareManagerSnippet.java
@@ -26,26 +26,34 @@ import com.google.android.mobly.snippet.rpc.Rpc;
 /** Snippet class exposing Android APIs in WifiAwareManager. */
 public class WifiAwareManagerSnippet implements Snippet {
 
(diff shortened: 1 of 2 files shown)


xpconanfan (changes requested):
> Reviewed 1 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/google/mobly-bundled-snippets/209)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @kaishi05)

___
*[`src/main/java/com/google/android/mobly/snippet/bundled/WifiAwareManagerSnippet.java` line 47 at r1](https://reviewable.io/reviews/google/mobly-bundled-snippets/209#-O4qL8se4LKPhyimR3mt:-O4qL8se4LKPhyimR3mu:b-e9hrso) ([raw file](https://github.com/google/mobly-bundled-snippets/blob/d3d513b919c07b321d9106a93e429c48dc6ded1b/src/main/java/com/google/android/mobly/snippet/bundled/WifiAwareManagerSnippet.java#L47)):*
> ```Java
>   public boolean wifiAwareIsAvailable() throws Throwable {
>     if (!mIsAwareSupported || mWifiAwareManager == null) {
>       return false;
> ```

we should throw an exception here instead.

Is available is clearly a different concept from `is supported`.

When it's not supported, `is available` is an invalid question, hence it should raise?

kaishi05:
As suggested, throw an exception when it is not supported.

src/main/java/com/google/android/mobly/snippet/bundled/WifiAwareManagerSnippet.java
```diff
@@ -26,26 +26,34 @@
...
+        if (mIsAwareSupported) {
+            mWifiAwareManager = (WifiAwareManager) mContext.getSystemService(Context.WIFI_AWARE_SERVICE);
+        }
+        Utils.adaptShellPermissionIfRequired(mContext);
+    }
+
+    /** Checks if Aware is available. This could return false if WiFi or location is disabled. */
+    @Rpc(description = "check if Aware is available.")
+    public boolean wifiAwareIsAvailable() throws WifiAwareManagerSnippetException {
+        if (!mIsAwareSupported || mWifiAwareManager == null) {
```
xpconanfan, line 53: one of these two is sufficient for determining whether wifi aware is supported on this device right?
can we simplify the logic and only have one signal var instead?

kaishi05: Done.

src/main/java/com/google/android/mobly/snippet/bundled/WifiAwareManagerSnippet.java
```diff
@@ -50,7 +49,7 @@ public WifiAwareManagerSnippet() throws Throwable {
     /** Checks if Aware is available. This could return false if WiFi or location is disabled. */
     @Rpc(description = "check if Aware is available.")
     public boolean wifiAwareIsAvailable() throws WifiAwareManagerSnippetException {
-        if (!mIsAwareSupported || mWifiAwareManager == null) {
+        if (mWifiAwareManager == null) {
```
xpconanfan, line 52: let's use `isAwareSupported` instead as it's more direct?

kaishi05: There is a small chance that mWifiAwareManager could be still null even if isAwareSupported = true. So I prefer the null check.

xpconanfan: But that null doesn't necessarily mean wifi aware is not supported.
If you are worried about that edge case, we should handle that separately with a different exception

kaishi05: Null means WifiAware service is not running and as the result Aware is not supported.

xpconanfan: "supported" means this device can support aware.
"available" means this device can perform wifi aware actions right now.

the edge case you brought up means Aware is supported on this device, but wifi aware manager being null
but that doesn't mean the wifi aware manager can't be null on this device ever?

kaishi05: Okay. If you were more clear on original suggestion, we would have less back and forth on this.  I made the change. Let's move on.

xpconanfan (approved):

PR #224, opened 2025-02-04
Update JsonDeserializer::jsonToBluetoothGattCharacteristic

AlanRosenthal:
Switch to using getIntBitwiseOr() for PROPERTY/PERMISSION

https://developer.android.com/reference/android/bluetooth/BluetoothGattCharacteristic

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/mobly-bundled-snippets/224)

### DIFF
--- a/src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonDeserializer.java
+++ b/src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonDeserializer.java
@@ -132,10 +132,10 @@ public class JsonDeserializer {
                         UUID.fromString(jsonObject.getString("UUID")),
                         // A characteristic can have multiple properties (e.g. PROPERTY_READ and PROPERTY_WRITE).
                         // Use the BitwiseOr to extract all the properties from the json string.
-                        MbsEnums.BLE_PROPERTY_TYPE.getIntBitwiseOr(jsonObject.getString("Property")),
+                        MbsEnums.BLE_PROPERTY_TYPE.getIntBitwiseOr(jsonObject.getString("Properties")),
                         // A characteristic can have multiple permissions (e.g. PERMISSION_READ and PERMISSION_WRITE).
                         // Use the BitwiseOr to extract all the permissions from the json string.
-                        MbsEnums.BLE_PERMISSION_TYPE.getIntBitwiseOr(jsonObject.getString("Permission")));
+                        MbsEnums.BLE_PERMISSION_TYPE.getIntBitwiseOr(jsonObject.getString("Permissions")));
         if (jsonObject.has("Data")) {
               dataHolder.insertData(characteristic, jsonObject.getString("Data"));
         }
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonDeserializer.java
```diff
@@ -130,8 +130,8 @@ public static BluetoothGattCharacteristic jsonToBluetoothGattCharacteristic(
         BluetoothGattCharacteristic characteristic =
                 new BluetoothGattCharacteristic(
                         UUID.fromString(jsonObject.getString("UUID")),
-                        MbsEnums.BLE_PROPERTY_TYPE.getInt(jsonObject.getString("Property")),
-                        MbsEnums.BLE_PERMISSION_TYPE.getInt(jsonObject.getString("Permission")));
+                        MbsEnums.BLE_PROPERTY_TYPE.getIntBitwiseOr(jsonObject.getString("Property")),
```
xpconanfan, line 133: can you explain why we need to use bitwise OR here instead?
perhaps with inline comments.

AlanRosenthal: added comment!

xpconanfan: Maybe use the full name of the example properties and permissions

Also, if it can be multiple, should we name them to "Properties" and "Permissions" instead?

AlanRosenthal: 1) I've updated the comments.
2) the "correct" names per the docs are "Properties" and "Permissions"
https://developer.android.com/reference/android/bluetooth/BluetoothGattCharacteristic#BluetoothGattCharacteristic(java.util.UUID,%20int,%20int)

However, changing this would break backwards compatibility with any existing application that uses "Property" and "Permission".

I can make the change, but I am not sure of the guarantee (if any) MBS provides to its users

AlanRosenthal: I thought about this some more and I think the best thing to do is to keep the name as is.

xpconanfan: I don't think there's a lot of, if any, usage of these fields today..

Question is, how is this represented in the Android API naming? Probably better to be consistent with that

AlanRosenthal: sounds good I updated to the API to match the Android API

xpconanfan (approved):

ko1in1u:
This PR will cause builder error, can we revert it?

```
src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonDeserializer.java:135: error: cannot find symbol
                        MbsEnums.BLE_PROPERTY_TYPE.getIntBitwiseOr(jsonObject.getString("Properties")),
                                                  ^
  symbol:   method getIntBitwiseOr(String)
  location: variable BLE_PROPERTY_TYPE of type RpcEnum

src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonDeserializer.java:138: error: cannot find symbol
                        MbsEnums.BLE_PERMISSION_TYPE.getIntBitwiseOr(jsonObject.getString("Permissions")));
                                                    ^
  symbol:   method getIntBitwiseOr(String)
  location: variable BLE_PERMISSION_TYPE of type RpcEnum
```

xpconanfan:
Feel free to revert or fix forward

PR #225, opened 2025-02-04
Include CCCD if requested (3/3)

AlanRosenthal:
This change is [Reviewable](https://reviewable.io/reviews/google/mobly-bundled-snippets/225)

### DIFF
--- a/src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonDeserializer.java
+++ b/src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonDeserializer.java
@@ -18,6 +18,7 @@ package com.google.android.mobly.snippet.bundled.utils;
 
 import android.annotation.TargetApi;
 import android.bluetooth.BluetoothGattCharacteristic;
+import android.bluetooth.BluetoothGattDescriptor;
 import android.bluetooth.BluetoothGattService;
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonDeserializer.java
```diff
@@ -135,6 +136,12 @@ public static BluetoothGattCharacteristic jsonToBluetoothGattCharacteristic(
         if (jsonObject.has("Data")) {
               dataHolder.insertData(characteristic, jsonObject.getString("Data"));
         }
+        if (jsonObject.has("CCCD") && jsonObject.getBoolean("CCCD")) {
```
AlanRosenthal, line 139: If the CCCD logic is too specific I can change this to genetic Descriptor logic (similar to the Characteristic logic).

xpconanfan: sorry I'm not familiar with this field.
could you share the API reference for this so we can make the right call?

In general making things more generic and matching the API design without implicit assumptions is recommended

AlanRosenthal: added generic support #239

PR #238, opened 2025-04-21
Update JsonSerializer.java to include BLE Identity Address

a111791a:
Reintroduce https://github.com/google/mobly-bundled-snippets/pull/236 guarded behind sdk version

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/mobly-bundled-snippets/238)

### DIFF
--- a/src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java
+++ b/src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java
@@ -134,8 +134,10 @@ public class JsonSerializer {
     public static Bundle serializeBluetoothDevice(BluetoothDevice data) {
         Bundle result = new Bundle();
         result.putString("Address", data.getAddress());
-        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.BAKLAVA) {
+        if (Build.VERSION.SDK_INT >= 36) {
             result.putString("IdentityAddress", data.getIdentityAddressWithType().getAddress());
+        } else {
+            result.putString("IdentityAddress", null);
         }
         final String bondState =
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java
```diff
@@ -134,7 +134,9 @@ private JSONObject serializeWifiInfo(WifiInfo data) throws JSONException {
     public static Bundle serializeBluetoothDevice(BluetoothDevice data) {
         Bundle result = new Bundle();
         result.putString("Address", data.getAddress());
-        result.putString("IdentityAddress", data.getIdentityAddressWithType().getAddress());
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.BAKLAVA) {
```
xpconanfan, line 137: we should probably put a None value for the same key when the SDK ver doesn't support this

xpconanfan, line 137: Is the version code BAKLAVA available on older SDK versioned phones?

xpconanfan (changes requested):

src/main/java/com/google/android/mobly/snippet/bundled/utils/JsonSerializer.java
ko1in1u: The Safe Method is using `Build.VERSION.SDK_INT`.

```
if (Build.VERSION.SDK_INT >= 36) {
   ...
}
```

a111791a:
Updated to compare Build.VERSION.SDK_INT with int directly and added default to getAddress() if sdk version < 36

ko1in1u (approved):

PR #240, opened 2025-05-02
Update compile SDK to 36 and add permission check when calling BluetoothDevice#getIdentityAddressWithType

ko1in1u:
This change is [Reviewable](https://reviewable.io/reviews/google/mobly-bundled-snippets/240)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -29,7 +29,7 @@ allprojects {
 apply plugin: 'com.android.application'
 
 android {
-    compileSdk 33
+    compileSdk 36
 
     defaultConfig {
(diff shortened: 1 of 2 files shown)


xpconanfan:
Would this incur new permission pop-ups in some tests..
Have we tested this internally?

xpconanfan (approved):

ko1in1u:
> Would this incur new permission pop-ups in some tests.. Have we tested this internally?

Yes, I've tested internally. Permission is checked at the API level, without any pop-up notifications on UI.

PR #242, opened 2025-07-29
Improve RPC WifiManagerSnippet#wifiScanAndGetResults.

mhaoli:
Changes:
1. Reference on the CTS-V Wi-Fi test impl, check EXTRA_RESULTS_UPDATED in intent and keep waiting if the most recent scan failed. 
2. Unregister WifiScanReceiver after completing the scan.

E2e tested locally.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/mobly-bundled-snippets/242)


mhaoli:
Converting this PR to draft for now as it cause some failures locally. 

According to the logs and API doc, the reason seems to be that one app cannot use trigger too many scans. Maybe not checking EXTRA_RESULTS_UPDATED makes sense for MBS. I'll check with Wi-Fi dev team and then decide how to handle the intent.

related logs:
```
07-30 23:12:30.533 I WifiScanRequestProxy: Scan request from com.google.snippet.wifi throttled
07-30 23:12:30.533 E WifiService: Failed to start scan
```

PR #246, opened 2025-09-03
Refactor `TelephonySnippet#getLine1Number` for improved reliability.

sashirestela:
Changes:

- Update getLine1Number to first attempt fetching the phone number for the specified simSlot using the active subscription instead of using the outdated TelephonyManager.getLine1Number() method.
- Add a fallback mechanism to retrieve the phone number from the default subscription if the slot-specific attempt yields a null or empty result.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/mobly-bundled-snippets/246)


sashirestela:
@ko1in1u This is the correct PR which has the CLA in place

PR #253, opened 2026-06-04
Keep adopted shell permission identity active in Mobly snippets

seanmahoney28-beep:
This PR removes the reflection-based call to UiAutomation.destroy().

Previously, the snippet attempted to manually tear down UiAutomation using reflection to allow other snippets to access global UI automation. However, destroying UiAutomation immediately also drops the shell permission identity that was just adopted.

By removing this manual teardown, the adopted shell permissions remain active so that subsequent snippet operations can continue to use them without issue.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/mobly-bundled-snippets/253)

### DIFF
--- a/src/main/java/com/google/android/mobly/snippet/bundled/utils/Utils.java
+++ b/src/main/java/com/google/android/mobly/snippet/bundled/utils/Utils.java
@@ -236,19 +236,7 @@ public final class Utils {
                         + Arrays.toString(permissions));
                 uia.adoptShellPermissionIdentity(permissions);
             }
-            // Need to drop the UI Automation to allow other snippets to get access
-            // to global UI automation.
-            // Using reflection here since the method is not public.
-            try {
-                Class<?> cls = Class.forName("android.app.UiAutomation");
-                Method destroyMethod = cls.getDeclaredMethod("destroy");
-                destroyMethod.invoke(uia);
-            } catch (NoSuchMethodException
-                    | IllegalAccessException
-                    | ClassNotFoundException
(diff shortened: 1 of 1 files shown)


seanmahoney28-beep (commented):
**Summary**
This PR removes the reflection-based call to `UiAutomation.destroy()`. 
Previously, the snippet attempted to manually tear down `UiAutomation` using reflection to allow other snippets to access global UI automation. However, destroying `UiAutomation` immediately also drops the shell permission identity that was just adopted. 
By removing this manual teardown, the adopted shell permissions remain active so that subsequent snippet operations can continue to use them without issue.
cc @xpconanfan

src/main/java/com/google/android/mobly/snippet/bundled/utils/Utils.java
```diff
@@ -236,19 +236,7 @@ public static void adaptShellPermissionIfRequired(
                         + Arrays.toString(permissions));
                 uia.adoptShellPermissionIdentity(permissions);
             }
-            // Need to drop the UI Automation to allow other snippets to get access
```
mhaoli, line 239: We need this fix since a recent UiAutomation behavior change. Thus let's keep this cleanup logic if SDK version < 37 to avoid affecting tests running on old devices. Thanks.

seanmahoney28-beep: This code change brought down all of our automation last week, happy to discard this change but please add a fix here thanks

PR #255, opened 2026-08-14
Add wifiDisconnect RPC to WifiManagerSnippet

moshenis:
Adds a `wifiDisconnect` RPC to `WifiManagerSnippet` so tests can drop the current Wi-Fi association without toggling the adapter off and on.

`WifiManager.disconnect()` returns `false` when the caller lacks the privileged permissions required on API 29+, so the return value is checked and a `WifiManagerSnippetException` is thrown on failure, matching the error handling used by the other RPCs in this class.

Tested on an API 35 emulator via the snippet server: the RPC is registered and returns successfully.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/mobly-bundled-snippets/255)

### DIFF
--- a/src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
+++ b/src/main/java/com/google/android/mobly/snippet/bundled/WifiManagerSnippet.java
@@ -96,6 +96,28 @@ public class WifiManagerSnippet implements Snippet {
             });
     }
 
+    /**
+     * Disconnects from the currently connected Wi-Fi network.
+     *
+     * <p>{@link WifiManager#disconnect()} is a privileged API: on API 29+ it is a no-op that
+     * returns {@code false} unless the snippet process holds the required system permissions.
+     */
+    @Rpc(description = "Disconnects from the currently connected Wi-Fi network with a 30s timeout.")
+    public void wifiDisconnect() throws InterruptedException, WifiManagerSnippetException {
+        if (!isWifiConnected()) {
+            return;
+        }
+        if (!mWifiManager.disconnect()) {
+            throw new WifiManagerSnippetException("Failed to initiate Wi-Fi disconnection.");
+        }
+        if (!Utils.waitUntil(() -> !isWifiConnected(), TIMEOUT_TOGGLE_STATE)) {
+            throw new WifiManagerSnippetException(
+                    String.format(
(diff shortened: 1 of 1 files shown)


xpconanfan (changes requested):
This might have been removed intentionally due to the "deprecated" designation.
So I'd like to understand your intended use case of this API.

In terms of the implementation:
1. In fact, this is a system API that only works when the snippet process has sufficient permissions. So it's probably worth noting that in comments
2.    `mWifiManager.disconnect()` only initiates the disconnection request asynchronously and returns `true` when the command is accepted. In tests, calling `wifiDisconnect()` followed immediately by an assertion or another action may cause race conditions if the connection hasn't finished tearing down.   Similar to `wifiEnable()`, `wifiDisable()`, and `wifiConnect()`, consider waiting for disconnection with `Utils.waitUntil`

moshenis:
Thanks for the review!

**Use case:** we have test scenarios that iterate over multiple Wi-Fi configurations — connect to network A, run assertions, then disconnect and connect to network B, and so on. Today the only way to drop the current association from MBS is `wifiDisable()` + `wifiEnable()`, which tears down the whole adapter. That is slow, and it also resets other Wi-Fi state that the test may care about, so the teardown between iterations is noisier than it needs to be. A direct `wifiDisconnect()` lets the test drop just the association.

I've updated the implementation to address both points:

1. Added a Javadoc note that `WifiManager.disconnect()` is a privileged API — on API 29+ it is a no-op returning `false` unless the snippet process holds the required system permissions.
2. `wifiDisconnect()` now returns early if Wi-Fi is not connected, and otherwise waits for the disconnection to actually complete via `Utils.waitUntil(() -> !isWifiConnected(), TIMEOUT_TOGGLE_STATE)`, consistent with `wifiEnable()`, `wifiDisable()` and `wifiConnect()`. It throws on timeout instead of returning while the teardown is still in flight.

Happy to adjust the timeout or the error messages if you'd prefer something different.

ko1in1u:
Consider using `wifiClearConfiguredNetworks` or modify this API to accept a specific ssid. Even `wifiDisconnect` works, the auto-connect feature will immediately reconnect to the Wi-Fi.

ko1in1u:
Or you can call `wifiConnect(networkB)` directly while connected to `networkA` without needing a separate disconnect step.

